PR #1, opened 2022-02-07
Add dependabot config

micgro42:
This config seems to have worked well for mismatch finder and is about to become our new standard.

Bug: T300883

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: monthly
+  open-pull-requests-limit: 10
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: monthly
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: monthly
+  open-pull-requests-limit: 10
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: monthly
+  open-pull-requests-limit: 15
```
lucaswerkmeister, line 12: Can we have a newline at the end of the file?

lucaswerkmeister (approved):

.github/dependabot.yml
micgro42: will do!

/me things about also adding an editor config

micgro42: done.

PR #2, opened 2022-02-07
Add package.json to allow adding dependencies

micgro42:
This should unlock of the follow-up tasks

Bug: T300883

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "new-lexeme-special-page",
+  "private": true,
+  "version": "0.0.1",
+  "description": "Form for the upcoming revival of the NewLexeme special page on Wikidata",
+  "main": "main.ts",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/wmde/new-lexeme-special-page.git"
+  },
+  "keywords": [
+    "Wikidata"
+  ],
+  "author": "The Wikidata team",
+  "license": "GPL-2.0-only",
```
lucaswerkmeister, line 18: Should it really be GPL2 *only*? I thought the default is or-later.

micgro42: Mh, our internal guideline doesn't say. I'll look at what we're using in other places.

micgro42: We don't seem to have this field in most `package.json` files. But the first WikibaeLexeme php file that I opened had a license tag for `GPL-2.0-or-later`, so I'm using that here as well

micgro42: Done.

lucaswerkmeister (approved):

PR #4, opened 2022-02-07
Add editorconfig to ensure newline at the end of every file

micgro42:


.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.{js,ts,vue}]
```
itamargiv, line 7: I'd say we'd probably want to add html to that as well, as we will have at least one html to actually test the page.

itamargiv (changes requested):
Nice work, one small comment

.editorconfig
lucaswerkmeister: Possibly JSON too? (Though the current `package.json` is indented with two spaces.)

micgro42: Done.

itamargiv (approved):
Thank you!

.editorconfig
micgro42: @lucaswerkmeister Sirry, didn't see your comment earlier. It is a good point and we probably want that for `.yml` as well. I'll address that in an extra PR that also fixes the two files

PR #5, opened 2022-02-07
Use tabs to indent yaml and json files

micgro42:
This is basically a follow-up to #4. Some form of linting for this will be added in the next PR.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,5 +4,8 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 
-[*.{js,ts,vue,html}]
+[*.{js,ts,vue,html,json}]
 indent_style = tab
+
+[*.{yml,yaml}]
+indent_size = 4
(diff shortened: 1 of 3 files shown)


lucaswerkmeister:
The dependabot YAML still seems to be indented with spaces as far as I can tell?

(Though for such a heavily-nested format, I’d also be fine with spaces, to be honest.)

micgro42:
> The dependabot YAML still seems to be indented with spaces as far as I can tell?
> 
> (Though for such a heavily-nested format, I’d also be fine with spaces, to be honest.)

Oh, you're right. Apparently, yaml cannot be indented with tabs: http://yaml.org/faq.html :thinking:

lucaswerkmeister:
Ah, that’s good to know :D

micgro42:
Ok, now `.yaml` should be consistently updated with **four** spaces and other files with tabs.

lucaswerkmeister (dismissed):
(Commit message is arguably outdated now, though.)

lucaswerkmeister (approved):

micgro42:
Fixed the commit message. Though I think, I maybe should undo that "Dismiss stale pull request approvals when new commits are pushed" option :thinking:

PR #6, opened 2022-02-07
Add husky setup and first pre-commit hook

micgro42:
This adds basic linting to yaml and json files.

As discussed, the pre-commit hook only checks files, it does not write anything. There is a `npm` script for that though.

Bug: T300893


lucaswerkmeister (approved):
Seems to work fine :+1:

PR #7, opened 2022-02-07
Lint js, ts and vue files

micgro42:
This also sets up scripts to fix these files and automatic linting in a pre-commit hook.

Bug: T300889

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,21 @@
+module.exports = {
+	root: true,
+	env: {
+		'vue/setup-compiler-macros': true,
+	},
+	extends: [
+		'wikimedia',
+		'wikimedia/node',
+		'wikimedia/client-common', // client-es6 still extends vue2 config
+		'wikimedia/vue3-es6',
+		'wikimedia/language/es2021', // this should ensure that we can use all the new featuresm, not only es6
+		'@wmde/wikimedia-typescript',
+		'@vue/typescript/recommended',
+	],
+	rules: {
+		'comma-dangle': [ 'error', 'always-multiline' ],
+
+		// resetting this back to the default value, to comply with modern vite conventions
+		'vue/component-tags-order': [ 'error', { order: [ [ 'script', 'template' ], 'style' ] } ],
+	},
+};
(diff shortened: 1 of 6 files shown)


lucaswerkmeister:
With `root: true` in the eslint config (see new commit) it works.

micgro42:
> With `root: true` in the eslint config (see new commit) it works.

Cool! +1 from my side to that extra commit

lucaswerkmeister (approved):
Do you know if it’s possible to move the ` --ext .js,.ts,.vue --ignore-path .gitignore` stuff to the eslintrc instead?

micgro42:
> > Do you know if it’s possible to move the ` --ext .js,.ts,.vue --ignore-path .gitignore` stuff to the eslintrc instead?
> 
> I'll have a look before I merge this

Meh, that feature seems to have received a no-fix in eslint/eslint#3529 😕

.eslintrc.js
```diff
@@ -0,0 +1,21 @@
...
+	root: true,
+	env: {
+		'vue/setup-compiler-macros': true,
+	},
+	extends: [
+		'wikimedia',
+		'wikimedia/node',
+		'wikimedia/client-common', // client-es6 still extends vue2 config
+		'wikimedia/vue3-es6',
+		'wikimedia/language/es2021', // this should ensure that we can use all the new featuresm, not only es6
```
lucaswerkmeister, line 11:
```suggestion
		'wikimedia/language/es2021', // this should ensure that we can use all the new features, not only es6
```

src/App.vue
```diff
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 // This starter template is using Vue 3 <script setup> SFCs
 // Check out https://v3.vuejs.org/api/sfc-script-setup.html#sfc-script-setup
-import HelloWorld from './components/HelloWorld.vue'
+import HelloWorld from './components/HelloWorld.vue';
```
lucaswerkmeister, line 4: Yup, I also noticed that this one was missing a semicolon in #9 :)

micgro42:
This now overrides another rule to allow for the default order of these top-level tags in SFC: 
```
'vue/component-tags-order': [ 'error', { order: [ [ 'script', 'template' ], 'style' ] } ],
```

lucaswerkmeister (commented):
Sorry, found one typo I must’ve missed earlier.

PR #9, opened 2022-02-07
Set up Wikit (vue3compat build)

lucaswerkmeister:
Use the migration build, mapping not only `vue` but also `@vue/composition-api` (used by Wikit) to `@vue/compat`. For `@click.native` to work, we need `MODE: 2` in the Vite config, or at least I haven’t yet found a way to make it work in `MODE: 3`. At some point, we’re hoping to have a properly Vue 3-compatible (subset of) Wikit, so that we can drop the migration build and use Vue 3 alone.

Bug: T301145

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,8 @@
 			"version": "0.0.1",
 			"license": "GPL-2.0-or-later",
 			"dependencies": {
+				"@vue/compat": "^3.2.30",
+				"@wmde/wikit-vue-components": "^2.1.0-alpha.11",
 				"vue": "^3.2.30"
 			},
 			"devDependencies": {
@@ -925,6 +927,14 @@
 			"integrity": "sha512-8TEXQxlldWAuIODdukIb+TR5s+9Ds40eSJrw+1iDDA9IFORPjMELarNQE3myz5XIkWWpdprmJjm1/SxMlWOC8A==",
 			"dev": true
 		},
+		"node_modules/@vue/compat": {
+			"version": "3.2.30",
+			"resolved": "https://registry.npmjs.org/@vue/compat/-/compat-3.2.30.tgz",
+			"integrity": "sha512-+k00eNjWNjpuAUQKB9kcGVQr6cTaXLu8bZ4cktiARZXjbnmk4QfdrJiRfayII+KGk3BfpFUIEkBcik/p+uV1Cg==",
+			"peerDependencies": {
+				"vue": "3.2.30"
+			}
+		},
 		"node_modules/@vue/compiler-core": {
 			"version": "3.2.30",
 			"resolved": "https://registry.npmjs.org/@vue/compiler-core/-/compiler-core-3.2.30.tgz",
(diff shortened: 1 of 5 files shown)


lucaswerkmeister:
Good question, I haven’t tried that yet either. But I think it should work.

lucaswerkmeister:
Eh, requires a bit more build setup work at least, Vite is currenty building ESM instead of CJS output.

lucaswerkmeister:
Alright, it works with the following changes here:
```diff
diff --git a/vite.config.ts b/vite.config.ts
index cb49666..f5071c1 100644
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -4,6 +4,22 @@ import vue from '@vitejs/plugin-vue';
 
 // https://vitejs.dev/config/
 export default defineConfig( {
+	build: {
+		target: 'esnext',
+		lib: {
+			entry: resolve( __dirname, 'src/main.ts' ),
+			name: 'main',
+			fileName: ( format ) => `SpecialNewLexeme.${format}.js`,
+		},
+		rollupOptions: {
+			output: {
+				globals: {
+					vue: 'Vue',
+				},
+			},
+			external: [ 'vue' ],
+		},
+	},
 	resolve: {
 		alias: {
 			'vue': '@vue/compat',
```

Plus the following changes in WikibaseLexeme, based on Noa’s special page patch:
```diff
diff --git a/WikibaseLexeme.resources.php b/WikibaseLexeme.resources.php
index 751f894b..9a55130a 100644
--- a/WikibaseLexeme.resources.php
+++ b/WikibaseLexeme.resources.php
@@ -254,6 +254,18 @@
 			"styles" => [
 				"lexeme.less"
 			]
-		]
+		],
+		"wikibase.lexeme.special.NewLexemeAlpha" => $moduleTemplate + [
+			"es6" => true,
+			"packageFiles" => [
+				'special/new-lexeme/dist/SpecialNewLexeme.umd.js',
+			],
+			"styles" => [
+				'special/new-lexeme/dist/style.css',
+			],
+			"dependencies" => [
+				"vue",
+			],
+		],
 	];
 } );
diff --git a/src/MediaWiki/Specials/SpecialNewLexemeAlpha.php b/src/MediaWiki/Specials/SpecialNewLexemeAlpha.php
index eae8ebd1..05fe0e71 100644
--- a/src/MediaWiki/Specials/SpecialNewLexemeAlpha.php
+++ b/src/MediaWiki/Specials/SpecialNewLexemeAlpha.php
@@ -31,8 +31,9 @@ public function execute( $subPage ): void {
 		$output = $this->getOutput();
 		$this->setHeaders();
 
-		$wikitext = 'Hello world!';
+		$wikitext = 'Hello world!<div id=app></div>';
 		$output->addWikiTextAsInterface( $wikitext );
+		$output->addModules( [ 'wikibase.lexeme.special.NewLexemeAlpha' ] );
 	}
 
 	public function setHeaders() {
```

But I’d do the build system changes separately.

lucaswerkmeister:
(Also, the above WikibaseLexeme diff assumes, for simplicity, that the dist files should be loaded from the submodule. This is of course not how we actually want to do it, I just didn’t set up the copying here.)

micgro42 (approved):
I think this is a step into the right direction for now. The next one will probably be about how to best externalize vue and create a hellow-world suitable for T300882

PR #10, opened 2022-02-07
Add some basic CI

micgro42:


.github/workflows/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+    js-test:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+
+            - name: Setup Node.js
+              uses: actions/setup-node@v2.5.1
+              with:
+                  node-version: "16.x"
```
lucaswerkmeister, line 13: (I’d prefer `"16"` so it’s the same string as in `.nvmrc` but that’s a very minor nitpick)

lucaswerkmeister (approved):

.github/workflows/test.yaml
micgro42: but that makes no difference in terms of which version will be chosen, right?

micgro42: done

lucaswerkmeister: Yeah, should be purely cosmetic.

PR #11, opened 2022-02-07
Setup stylelint

itamargiv:
This change sets up stylelint, as well as reorganizes some of our npm scripts for consistency and ease of use within CI workflows.

Bug: [T300888](https://phabricator.wikimedia.org/T300888)


package.json
```diff
@@ -10,10 +10,15 @@
...
 		"test": "npm-run-all test:*",
-		"test:lint": "eslint --ext .js,.ts,.vue --ignore-path .gitignore .",
-		"fix": "npm-run-all fix:*",
+		"test:unit": "echo 'Tests TBD'",
+		"lint": "run-s lint:*",
+		"lint:config": "prettier --check '**/*.{json,yml,yaml}'",
+		"lint:script": "eslint --ext .js,.ts,.vue --ignore-path .gitignore .",
+		"lint:style": "stylelint '**/*.{css,scss,vue}'",
+		"fix": "run-s fix:*",
+		"fix:config": "prettier --write '**/*.{json,yml,yaml}'",
```
micgro42, line 19: While I understand the motivation here, we do also have some config in `.js` and `.ts` files, that is why I went for mentioning the file extensions for this particular line explicitly.

itamargiv: I'm not sure why it particularly matters in this case (these types of files will almost always hold config while ts and js are more versatile, just as vue files also hold styles), but if you insist I will change it back.

micgro42: I'm thinking about it from the other direction. If I want to fix spacing in my config files, then running `npm run fix:config` maybe won't be enough, even though that command claims to fix the config. Whereas the command `npm run fix:jsonYaml` is more honest about what it is doing. That being said, I don't like the name `fix:jsonYaml` much either and am still open to improvements.

Also, I very much appreciate your work to restructure our scripts, and am grateful that you added that linting script for json and yaml files that I had forgotten :pray:

lucaswerkmeister:
I think linting should be part of the `test` command – if it’s in a separate command, then I’ll probably just forget to run it and get annoyed about pushing bad commits all the time.

package.json
lucaswerkmeister: I feel like we might as well name them after the tool: `fix:prettier`, `fix:eslint`, `fix:stylelint`. I’m not sure I would run any of these individually anyways – the `lint-staged` approach seems more convenient (but is there something like `fix-staged`?).

lucaswerkmeister:
I don’t think that’s such a clear separation. We’ll have several kinds of tests anyways – unit tests, browser tests, maybe e2e tests, maybe other kinds of tests. And at least eslint can also detect some logic mistakes, like typoed variable names. A TypeScript compilation check could also be another kind of test, which I think doesn’t really fit into this test/lint split.

I also don’t really see a situation where I would want to run all the different kinds of test, but not the lint commands… I think I’d either want to run only one kind of test (e.g. `npm run test:unit` or even `npm run test:unit tests/unit/components/Foo.spec.ts`), or run all the checks we have (including lints, i.e. same as CI).

itamargiv:
@lucaswerkmeister I've added a `check` script as a compromise

itamargiv:
> I also don’t really see a situation where I would want to run all the different kinds of test, but not the lint commands… I think I’d either want to run only one kind of test (e.g. `npm run test:unit` or even `npm run test:unit tests/unit/components/Foo.spec.ts`), or run all the checks we have (including lints, i.e. same as CI).

I actually do that quite often, maybe this codebase is not so bloated yet, but yes, I'd like to maintain a separation between the two.

package.json
itamargiv: > I'm thinking about it from the other direction. If I want to fix spacing in my config files, then running npm run fix:config maybe won't be enough, even though that command claims to fix the config.

Very good point by the way

lucaswerkmeister:
> I actually do that quite often, maybe this codebase is not so bloated yet, but yes, I'd like to maintain a separation between the two.

Which command(s) do you often run – test, lint, or both?

itamargiv:
> Which command(s) do you often run – test, lint, or both?

Full test suite w/o lint (especially if I have to handle some sore of behemoth of vue + typescript + es/tslint)

itamargiv:
Eventually it was settled in mattermost, that we shall use `test` for the sake of convenience npm shorthands such as `npm t` or `npm it`.

lucaswerkmeister (dismissed):
I’m vaguely tempted to sort the npm scripts alphabetically but that can really wait :D

lucaswerkmeister (approved):

PR #14, opened 2022-02-08
Add jest scaffolding

micgro42:
The test currently imports the local component with a relative link
instead of an alias because of [T301267](https://phabricator.wikimedia.org/T301267).

Co-authored-by: Lucas Werkmeister <lucas.werkmeister@wikimedia.de>
Bug: [T300890](https://phabricator.wikimedia.org/T300890)

PR #18, opened 2022-02-09
Configure eslint-plugin-import to lint imports

lucaswerkmeister:
This replaces eslint-plugin-node’s no-missing-import rule, which doesn’t support `@/…`-style aliases we want to use.

Bug: T301267


micgro42:
Looks good!

Could you also change the import in our one test file? That is also from #14 which had spawned this issue.

```
 import HelloWorld from '../../src/components/HelloWorld.vue';
```

PR #19, opened 2022-02-10
Prepare integration with WikibaseLexeme

lucaswerkmeister:
See commit messages. (Could also be split into several PRs, not sure.)

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,6 +7,10 @@
 	</head>
 	<body>
 		<div id="app"></div>
-		<script type="module" src="/src/main.ts"></script>
+		<script type="module">
+			import createAndMount from './src/main.ts';
+
+			createAndMount( { rootSelector: '#app' } );
+		</script>
 	</body>
 </html>
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,11 +1,14 @@
-import { createApp } from 'vue';
+import {
+	ComponentPublicInstance,
+	createApp,
+} from 'vue';
 import App from './App.vue';
 
 interface Config {
 	rootSelector: string;
 }
 
-export default function createAndMount( config: Config ): void {
-	createApp( App )
+export default function createAndMount( config: Config ): ComponentPublicInstance {
+	return createApp( App )
 		.mount( config.rootSelector );
 }
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,4 +1,11 @@
 import { createApp } from 'vue';
 import App from './App.vue';
 
-createApp( App ).mount( '#app' );
+interface Config {
+	rootSelector: string;
+}
+
+export default function createAndMount( config: Config ): void {
+	createApp( App )
+		.mount( config.rootSelector );
+}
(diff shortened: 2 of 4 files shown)


vite.config.ts
```diff
@@ -6,6 +6,18 @@ import vue from '@vitejs/plugin-vue';
 
 // https://vitejs.dev/config/
 export default defineConfig( {
+	build: {
+		target: 'esnext',
```
itamargiv, line 10: :+1: I really like that we can work with it now in mw code

index.html
```diff
@@ -7,6 +7,10 @@
 	</head>
 	<body>
 		<div id="app"></div>
-		<script type="module" src="/src/main.ts"></script>
+		<script type="module">
+			import createAndMount from './src/main.ts';
+
+			createAndMount( { rootSelector: '#app' } );
```
itamargiv, line 13: yes, pretty much what I had in mind +1 Thank you

itamargiv (dismissed):
Looks good and straightforward, no obvious pitfalls when building locally. Thank you!

lucaswerkmeister:
Here’s the WikibaseLexeme half of this: https://gerrit.wikimedia.org/r/c/mediawiki/extensions/WikibaseLexeme/+/761679

vite.config.ts
lucaswerkmeister: Can we, actually? :sweat_smile: I think I just copied that from somewhere, it would probably be smart to check if it’s correct or not. We still support some other old-ish browsers, after all (Proxies aren’t _that_ new). I’ll try to look into that later.

itamargiv: oops, I suppose we can't :P

itamargiv: https://github.com/wikimedia/design-codex/blob/b76f6d26389618ac266fb72f5747b157115ef78d/packages/codex/vite.config.ts#L10

lucaswerkmeister: Thanks! Though after looking at the rest of wikimedia/design-codex@f9614ae76e4e974755a2e9a1a9927232033cf810, I notice they add `es2015` in some other places too, and we also already have `esnext` in our `.tsconfig` and should probably fix that. I guess that can be another pull request before this one.

lucaswerkmeister: #23

lucaswerkmeister:
Rebased onto #21, #22 and #23 – should be ready to merge now, I think.

src/main.ts
```diff
@@ -1,4 +1,11 @@
 import { createApp } from 'vue';
 import App from './App.vue';
 
-createApp( App ).mount( '#app' );
+interface Config {
+	rootSelector: string;
+}
+
+export default function createAndMount( config: Config ): void {
```
itamargiv, line 8: I didn't notice it until now, but I think it might be more convenient to work with, if we return the mounted app as well.

itamargiv (commented):
One small comment, and then I think we're good to go

src/main.ts
lucaswerkmeister: Done.

micgro42: Out of curiosity: what would we do with the mounted app on-wiki?

PR #21, opened 2022-02-11
Define typeRoots in tsconfig.json

lucaswerkmeister:
We plan to use this repository as a submodule of WikibaseLexeme.git, but we don’t want TypeScript to use WikibaseLexeme’s node_modules/@types. (Not only are they not useful to us, they even cause errors for some reason that I cannot be bothered to investigate.)

Co-Authored-By: Michael Große <michael.grosse@wikimedia.de>


quetzlthethird (approved):
looks good :)

PR #23, opened 2022-02-11
Change esnext to es2015

lucaswerkmeister:
Following the Codex design system, which probably targets the same set of browsers we target, and which we want to use at some point anyways.


itamargiv (approved):
Thank you!

PR #26, opened 2022-02-11
Stop using eslint-plugin-import

lucaswerkmeister:
It’s not correctly understanding our imports yet, producing unwanted errors, and even if we put in the time to configure it correctly, its usefulness seems dubious at best. (If we really have broken imports, we’d probably find out about it during the build or test steps.) Let’s just remove it.


itamargiv (approved):
:+1: :

PR #28, opened 2022-02-11
WIP add Lemma input

micgro42:


src/components/LemmaInput.vue
```diff
@@ -0,0 +1,32 @@
...
+
+const $messages = inject( MessagesKey, new Messages() );
+</script>
+
+<template>
+	<text-input
+		:label="$messages.get( 'wikibaselexeme-newlexeme-lemma' )"
+		:placeholder="$messages.get( 'wikibaselexeme-newlexeme-lemma-placeholder' )"
+		name="lemma"
+		required
```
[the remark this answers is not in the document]
lucaswerkmeister: Should we wait with merging this until the new release is out?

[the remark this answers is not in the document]
itamargiv: We can also just make a pre-release :D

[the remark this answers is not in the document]
micgro42: I think everything that we need from wikit should be currently in the master branch over there, including the tokens for the License text.

src/components/LemmaInput.vue
```diff
@@ -0,0 +1,31 @@
+<script setup lang="ts">
+import { TextInput } from '@wmde/wikit-vue-components';
+import { useMessages } from '@/plugins/MessagesPlugin/Messages';
+
+withDefaults( defineProps<{
+	modelValue: string;
+	isErrorInputTooLong: boolean;
```
itamargiv, line 7: Not blocking, and probably just aesthetic, but I find this prop name to be a bit verbose (and perhaps extending this to support more errors could be a bit more cumbersome in the future). How about renaming this to `errors` and passing down an object or array to flag the errors to display?

tests/unit/components/LemmaInput.test.ts
```diff
@@ -0,0 +1,44 @@
...
+			},
+			...config,
+		} );
+	}
+
+	function findInput( wrapper: ReturnType<typeof createComponent> ): Omit<DOMWrapper<HTMLInputElement>, 'exists'> {
+		return wrapper.get( 'input' );
+	}
+
+	describe( ':props', () => {
```
itamargiv, line 19: I really like this testing structure, never seen it before, but definitely adopted!

itamargiv (commented):
Great tests make great work! :100: I only have one qualm / question, but it's mostly stylistic, so not blocking really...

tests/unit/components/LemmaInput.test.ts
micgro42: I glimpsed it in a talk about vue testing and really liked it a lot as well: https://www.youtube.com/watch?v=OIpfWTThrK8
No idea how found this talk, though. It is from way before I had ever looked at vue^^

src/components/LemmaInput.vue
micgro42: ~~I considered that and tried to weigh the tradeoffs. T298137 only specifies this single error, Figma does have copy for a different error, "No input at all", but that should never come to pass, because the `required` attribute wouldn't even allow the form to be submitted. So adding complexity beyond the simplest thing to do for this single-purpose component could be YAGNI?~~

And I just realized that the `required` attribute can be circumvented by only typing a space, which we probably don't want to allow. Thus, we need to support at least two errors indeed. I'll update the component. Thanks for making me think about this again. :+1:

src/components/LemmaInput.vue
```diff
@@ -0,0 +1,43 @@
+<script setup lang="ts">
+import { TextInput } from '@wmde/wikit-vue-components';
+import { useMessages } from '@/plugins/MessagesPlugin/Messages';
+
+interface Props {
+	modelValue: string;
+	error?: 'ERROR_LEMMA_TOO_LONG' | 'ERROR_NO_LEMMA' | null;
```
itamargiv, line 7: I like this error key approach, we can definitely develop this further from here. Thank you!

src/components/LemmaInput.vue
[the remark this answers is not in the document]
lucaswerkmeister: [2.1.0-alpha.12](https://www.npmjs.com/package/@wmde/wikit-vue-components/v/2.1.0-alpha.12) exists now \o/

src/components/LemmaInput.vue
micgro42: Now looking at actually implementing errors, I'm no longer sure this was the best idea^^
At the very least it was too early to do it in this particular pull request

PR #29, opened 2022-02-11
Add i18n plugin

micgro42:
In the compositionApi we cannot access `this`, so using a global instance method does not work. The established way to deal with this in Vue 3 is to use provide/inject.

Using provide/inject works well, with the one exception that requires a fallback or will be typed `Messages|undefined`. This leaves a couple of non-ideal solutions:
1. assert it as non-null it with `!`
2. cast it with `as Messages`
3. have some if-guard that throws an Error
4. Provide a simple fallback

I went with option 4 and extended the original Messages class from [Data Bridge] with a fallback if no repository is provided. This also makes the initialization in the index.html easier and still allows for a more elaborate mock in the future, if we want it.

[Data Bridge]: https://github.com/wikimedia/Wikibase/blob/f73e98507da6e494ffaa1fe9c15b787a527ba33a/client/data-bridge/src/presentation/plugins/MessagesPlugin/Messages.ts

Bug: T301570

### DIFF
--- a/src/components/HelloWorld.vue
+++ b/src/components/HelloWorld.vue
@@ -3,20 +3,24 @@
 <script setup lang="ts">
 import { useStore } from 'vuex';
 import { Button as WikitButton } from '@wmde/wikit-vue-components';
-import { computed } from 'vue';
+import { computed, inject } from 'vue';
 import { DELAYED_INCREMENT } from '@/store/actions';
+import Messages, { MessagesKey } from '@/plugins/MessagesPlugin/Messages';
 
 defineProps<{ msg: string }>();
 
 const store = useStore();
 const count = computed( () => store.state.count );
 const increment = () => store.dispatch( DELAYED_INCREMENT, { delay: 300 } );
+const $messages = inject( MessagesKey, new Messages() );
 </script>
 
 <template>
 	<h1>{{ msg }}</h1>
 
 	<p>See <code>README.md</code> for more information.</p>
+
+	<p>i18n example: {{ $messages.get( 'wikibaselexeme-newlexeme-lemma' ) }}</p>
 
 	<p>
 		<a href="https://vitejs.dev/guide/features.html" target="_blank">
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,13 +4,21 @@ import {
 } from 'vue';
 import store from './store';
 import App from './App.vue';
+import Messages, { MessagesKey } from './plugins/MessagesPlugin/Messages';
+import MessagesRepository from './plugins/MessagesPlugin/MessagesRepository';
 
 interface Config {
 	rootSelector: string;
 }
 
-export default function createAndMount( config: Config ): ComponentPublicInstance {
-	return createApp( App )
-		.use( store )
-		.mount( config.rootSelector );
+export default function createAndMount(
+	config: Config,
+	messageRepo?: MessagesRepository,
+): ComponentPublicInstance {
+	const app = createApp( App );
+	app.use( store );
+
+	app.provide( MessagesKey, new Messages( messageRepo ) );
+
+	return app.mount( config.rootSelector );
 }
--- a/src/plugins/MessagesPlugin/MessageKeys.ts
+++ b/src/plugins/MessagesPlugin/MessageKeys.ts
@@ -1,0 +1,4 @@
+type MessageKeys =
+	'wikibaselexeme-newlexeme-lemma';
+
+export default MessageKeys;
--- a/src/plugins/MessagesPlugin/Messages.ts
+++ b/src/plugins/MessagesPlugin/Messages.ts
@@ -7,8 +7,8 @@ import { InjectionKey } from 'vue';
  *
  * FIXME: write something about when to use getText vs get
  *
- * `this.$messages.get( this.$messages.KEYS.LEMMA_INPUT_LABEL )`
- * `this.$messages.getText( this.$messages.KEYS.LEMMA_INPUT_LABEL )`
+ * `this.$messages.get( 'wikibaselexeme-newlexeme-lemma' )`
+ * `this.$messages.getText( 'wikibaselexeme-newlexeme-lemma' )`
  */
 export default class Messages {
 
--- a/src/plugins/MessagesPlugin/Messages.ts
+++ b/src/plugins/MessagesPlugin/Messages.ts
@@ -1,0 +1,41 @@
+import MessagesRepository from '@/plugins/MessagesPlugin/MessagesRepository';
+import MessageKeys from '@/plugins/MessagesPlugin/MessageKeys';
+import { InjectionKey } from 'vue';
+
+/**
+ * Usage (assuming this has been registered as a Vue plugin):
+ *
+ * FIXME: write something about when to use getText vs get
+ *
+ * `this.$messages.get( this.$messages.KEYS.LEMMA_INPUT_LABEL )`
+ * `this.$messages.getText( this.$messages.KEYS.LEMMA_INPUT_LABEL )`
+ */
+export default class Messages {
+
+	private readonly messagesRepository: MessagesRepository;
+
+	public constructor( messagesRepository?: MessagesRepository ) {
+		if ( !messagesRepository ) {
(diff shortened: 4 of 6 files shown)


src/components/HelloWorld.vue
```diff
@@ -1,17 +1,21 @@
 <script setup lang="ts">
-import { ref } from 'vue';
+import { ref, inject } from 'vue';
 import { Button as WikitButton } from '@wmde/wikit-vue-components';
+import Messages, { MessagesKey } from '@/plugins/MessagesPlugin/Messages';
 
 defineProps<{ msg: string }>();
 
 const count = ref( 0 );
+const $messages = inject( MessagesKey, new Messages() );
```
itamargiv, line 9: I like that we have adopted the composition API in full. This is a great use case example of provide and inject.

src/plugins/MessagesPlugin/Messages.ts
```diff
@@ -0,0 +1,43 @@
...
+			return;
+		}
+		this.messagesRepository = messagesRepository;
+	}
+
+	public get( messageKey: string, ...params: readonly ( string|HTMLElement )[] ): string {
+		return this.messagesRepository.get( messageKey, ...params );
+	}
+
+	public getUnescaped( messageKey: string, ...params: readonly string[] ): string {
```
itamargiv, line 38: Unescaped in this context means the raw text string? What is the use case example for unescaped text if we don't plan on using it as `innerHTML`?

itamargiv (commented):
This looks excellent. I have one question, just to clarify the usage context of the `MessagesRepository` interface, but in any case, this is a great adapter :+1:

src/plugins/MessagesPlugin/Messages.ts
micgro42: We want to use this in templates directly with `{{}}`, e.g. `<label>{{ messages.getUnescaped( 'label-text ) }}</label>` where it would be double-escaped with just using `.get()`.

itamargiv: Aha, I see, thank you!

src/components/HelloWorld.vue
```diff
@@ -1,17 +1,21 @@
...
 </script>
 
 <template>
 	<h1>{{ msg }}</h1>
 
 	<p>See <code>README.md</code> for more information.</p>
 
+	<p>i18n example: {{ $messages.get( $messages.KEYS.LEMMA_INPUT_LABEL ) }}</p>
```
lucaswerkmeister, line 17: This discussion doesn’t have to block merging, but I wonder what the benefit of the `KEYS` enum is here? I think it would be easier to find the message used here in `i18n/en.json`, or to find all the places in the code using a certain message, if we removed this level of indirection and just had:

```html
 	<p>i18n example: {{ $messages.get( 'wikibaselexeme-newlexeme-lemma' ) }}</p>
```

We could still have tsc check that only known message keys are used by declaring the message key as `keyof MessageKeys` or something like that.

lucaswerkmeister (dismissed):

src/main.ts
```diff
@@ -3,12 +3,20 @@ import {
...
 interface Config {
 	rootSelector: string;
 }
 
-export default function createAndMount( config: Config ): ComponentPublicInstance {
-	return createApp( App )
-		.mount( config.rootSelector );
+export default function createAndMount(
+	config: Config,
+	messageRepo?: MessagesRepository,
```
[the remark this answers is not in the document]
itamargiv: I think we need to distinguish between injecting configurations and injecting services. Config for example is probably also going to receive a URL for the form to be sent to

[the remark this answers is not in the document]
micgro42: To me, I think I would treat "config" and "services" differently. But this is not intended to be the final form of that signature yet, I would probably collect the `messageRepo` and `mw.api` into one object

lucaswerkmeister:
Since #27 was just merged, it would be nice to have a unit test for the `provide()` too – is that possible in VTU?

src/components/HelloWorld.vue
micgro42: Yes, I like that more, and we even still have IDE autocompletion! 🙌

src/plugins/MessagesPlugin/Messages.ts
```diff
@@ -0,0 +1,41 @@
...
+import MessageKeys from '@/plugins/MessagesPlugin/MessageKeys';
+import { InjectionKey } from 'vue';
+
+/**
+ * Usage (assuming this has been registered as a Vue plugin):
+ *
+ * FIXME: write something about when to use getText vs get
+ *
+ * `this.$messages.get( this.$messages.KEYS.LEMMA_INPUT_LABEL )`
+ * `this.$messages.getText( this.$messages.KEYS.LEMMA_INPUT_LABEL )`
```
lucaswerkmeister, line 11:
```suggestion
 * `this.$messages.get( 'wikibaselexeme-newlexeme-lemma' )`
 * `this.$messages.getText( 'wikibaselexeme-newlexeme-lemma' )`
```

micgro42:
> Since #27 was just merged, it would be nice to have a unit test for the `provide()` too – is that possible in VTU?

I have not yet found a way to do that. Though, I'm having trouble inspecting the object returned by `createAndMount()`. It seems to be a Proxy of a Proxy? o.O

src/plugins/MessagesPlugin/MessageKeys.ts
```diff
@@ -0,0 +1,4 @@
+type MessageKeys =
+	'wikibaselexeme-newlexeme-lemma';
```
itamargiv, line 2: Wait, isn't `MessageKeys` used to annotate types for the plugin's `get()` and `getUnescaped`? I'm not sure a literal alias is the way to go here, unless I'm missing something

micgro42: Mh, I maybe should have added more. My goal is to extend this as a list of all allowed messages:

```ts
type MessageKeys =
	'wikibaselexeme-newlexeme-lemma'
	| 'foo'
	| 'bar';

```

micgro42: That is, the type is intended to be "one of these exact literal strings"

micgro42: > Ah, yeah I got that, but I thought this would be derived from an object similar to what we added in PR #33 (exported from a file ofc, and not just ad-hoc in index.html):
> 
> https://github.com/wmde/new-lexeme-special-page/blob/5492aa7ddae6ec22b56d8d499f650d604fb5d5e2/index.html#L13-L15

But, these messages are being stored outside of this git repository, in https://github.com/wikimedia/mediawiki-extensions-WikibaseLexeme/blob/master/i18n/en.json

I don't think duplicating this copy is a good idea. They are bound to diverge.

itamargiv: Yes, I also understand that. I just think having a version of the copy available for local development could be useful, especially for visual tasks like copy on a button. It doesn't have to stay in sync.

micgro42: Yeah, having mocks for the copy makes sense, I agree with that. But mocks are no source for types for production code. Rather the other way around, the mock could be typed with the production types:

```ts
 const mockMessages: Record<MessageKeys, string> = { 
 	'wikibaselexeme-new-lexeme-submit': 'Create Lexeme', 
 };
 ```
 
 or something like that.

src/main.ts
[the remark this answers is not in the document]
lucaswerkmeister: Alright, I guess we can do that later.

src/plugins/MessagesPlugin/MessageKeys.ts
itamargiv: > But mocks are no source for types for production

Agreed, just wanted to prevent having to synchronize the same keys everywhere.

lucaswerkmeister (approved):
LGTM, should be enough to unblock #33 even if we change the details later.

itamargiv (approved):

PR #30, opened 2022-02-11
Upgrade stylelint-config-wikimedia to resolve vuln notification

micgro42:
With this we can use a more recent version of postcss that is no longer vulnerable and that yellow bar should go away.


itamargiv (approved):
Thanks for the upgrade! I think we can ignore the failed netlify deploy on this one.

lucaswerkmeister (approved):

PR #31, opened 2022-02-14
fix Netlify hosting

micgro42:
Netlify actually needs an index.html in the dist directory. However this
is not added for a library build. So this creates an extra build command
to keep building it as an app that makes still use of the index.html,
link `npm run dev`. This skips the type checking as the build currently
runs on Netlify itself and there we only have a limited amount of build
minutes included. Rather the type checking should be part of the "code
quality" CI job.

This Netlify hosting was broken in #19 / 171d33f

* [ ] Open question: Do we want this to be built in `--mode development`? 
This would allow devtools to work, but might in theory have slight differences
to production?
* [ ] Also: this will need adjustments to the config in Netlify to work. (that is, to the build command)

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,12 +1,17 @@
 /* eslint quote-props: [ "error", "consistent-as-needed" ] */
 
 import { resolve } from 'path';
-import { defineConfig } from 'vite';
+import { BuildOptions, defineConfig } from 'vite';
 import vue from '@vitejs/plugin-vue';
 
-// https://vitejs.dev/config/
-export default defineConfig( {
-	build: {
+function getBuildConfig( isAappBuild: bool ): BuildOptions {
+	if ( isAappBuild ) {
+		return {
+			target: 'es2015',
+		};
+	}
+
+	return {
 		target: 'es2015',
 		lib: {
 			entry: resolve( __dirname, 'src/main.ts' ),
@@ -17,7 +22,12 @@ export default defineConfig( {
 		rollupOptions: {
 			external: [ '@vue/compat', 'vue', 'vuex' ],
 		},
-	},
+	};
+}
+
+// https://vitejs.dev/config/
+export default defineConfig( {
+	build: getBuildConfig( !!process.env.BUILD_AS_APP ),
 	resolve: {
 		alias: {
 			'vue': '@vue/compat',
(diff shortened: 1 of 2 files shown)


vite.config.ts
```diff
@@ -1,12 +1,17 @@
...
 import { resolve } from 'path';
-import { defineConfig } from 'vite';
+import { BuildOptions, defineConfig } from 'vite';
 import vue from '@vitejs/plugin-vue';
 
-// https://vitejs.dev/config/
-export default defineConfig( {
-	build: {
+function getBuildConfig( isAappBuild: bool ): BuildOptions {
+	if ( isAappBuild ) {
```
lucaswerkmeister, line 8:
```suggestion
function getBuildConfig( isAppBuild: bool ): BuildOptions {
	if ( isAppBuild ) {
```

vite.config.ts
```diff
@@ -17,7 +22,12 @@ export default defineConfig( {
 		rollupOptions: {
 			external: [ '@vue/compat', 'vue', 'vuex' ],
 		},
-	},
+	};
+}
+
+// https://vitejs.dev/config/
+export default defineConfig( {
+	build: getBuildConfig( !!process.env.BUILD_AS_APP ),
```
lucaswerkmeister, line 30: Optional: maybe we should `import { env } from 'process';`?

lucaswerkmeister (changes requested):

vite.config.ts
micgro42: 🤦

Thanks!

lucaswerkmeister (approved):

vite.config.ts
micgro42: I think I prefer `process.env` as it is such an established pattern and it makes it obvious right away that it is coming from the environment variables directly, without needing to scroll up and see where `env` comes from.

lucaswerkmeister: We could still have that with `import process from 'process'` if I’m not mistaken – to me it feels weird how Node has some of these modules that you can `import`/`require` but also they just exist by default for some reason. ([The docs](https://nodejs.org/api/process.html#process) say that the explicit import is recommended, apparently.)

lucaswerkmeister (approved):

PR #32, opened 2022-02-14
Provide Branch Hosting for GitHub repo with Netlify

guergana:
Bug: [T301446](https://phabricator.wikimedia.org/T301446)


README.md
```diff
@@ -11,3 +11,12 @@ _**Prerequisits: This repository requires node v16 and up**_
...
 3. Test and lint with `npm test`
+
+## Manual Deployment
+
+Before making a deploy make sure you make a build locally of the content that should be deployed by running `npm run buildPreview`.
+
+Afterwards, run `npm run netlify deploy` to run manual deployments to a unique address that will be generated at the time of deploy. 
+To deploy to [https://new-lexeme-special-page.netlify.app](https://new-lexeme-special-page.netlify.app) run `npm run netlify deploy --prod`. 
+
+To be able to run the netlify commands you will need to run `netlify login` once after installing netlify to authenticate and obtain a token.
```
lucaswerkmeister, line 22: What does installing netlify mean here? Is it part of the `npm install` or something else? (And if it is part of the `npm install`, shouldn’t the command be `npm run netlify login`, so it uses the `netlify` command from `node_modules`?)

guergana: ah yes, it should!!! my bad. i will fix it. installing netlify means running `npm install netlify-cli --save-dev`

guergana: Done, the command is updated!

README.md
```diff
@@ -11,3 +11,12 @@ _**Prerequisits: This repository requires node v16 and up**_
 1. Clone this repository
 2. Install dependecies with `npm i`
 3. Test and lint with `npm test`
+
+## Manual Deployment
+
+Before making a deploy make sure you make a build locally of the content that should be deployed by running `npm run buildPreview`.
+
+Afterwards, run `npm run netlify deploy` to run manual deployments to a unique address that will be generated at the time of deploy.
```
lucaswerkmeister, line 19: It’s asking me for a publish directory. What’s the right option – `dist` or `.`?

package.json
```diff
@@ -12,6 +12,7 @@
 		"fix:prettier": "prettier --write '**/*.{json,yml,yaml}'",
 		"fix:eslint": "eslint --fix --ext .js,.ts,.vue --ignore-path .gitignore . --ignore-pattern 'cypress'",
 		"fix:stylelint": "stylelint --fix '**/*.{css,scss,vue}'",
+		"netlify": "netlify",
```
lucaswerkmeister, line 15: I wonder if this is worth adding… I’d probably just use `npx netlify` instead of `npm run netlify` most of the time, since it’s shorter. Do we want to put that in the README instead?

README.md
guergana: dist.

package.json
guergana: what would the difference be between the two? :eyes:

README.md
guergana: do you think I should add that information to the docs??? i guess that can also be inferred from looking at what buildPreview does, but I can also make it clearer here.

lucaswerkmeister: I definitely think it should be added to the docs, yes.

package.json
lucaswerkmeister: No difference, as far as I’m aware. One’s just shorter than the other.

guergana: if I don't add the script i can't run netlify with npm, unless i specify the full path to the bin in `node_modules`. I would prefer to leave the npm command because i think it is more used that `npx` and easier to understand for other people looking at the code

lucaswerkmeister: Well, you can also use `npm x netlify` – `npx` is just a shortcut for `npm x` (= `npm exec`), if I understand correctly. But if you want to keep this one, I don’t mind either.

README.md
guergana: i added it.

package.json
guergana: Yes, if the user knows how to use the `npx` command I guess they can also figure that one out and decide if they want to use `npm` or `npx`. [I see that a difference is that with npx](https://www.freecodecamp.org/news/npm-vs-npx-whats-the-difference/) you can use packages that are not installed locally, but we are installing netlify locally.

README.md
lucaswerkmeister: Hm, I’m not sure that would answer my question… I meant adding it to the paragraph about `npm run netlify deploy`, since that is the command asking about the publish directory.

lucaswerkmeister: Maybe something like:

> Afterwards, run `npm run netlify deploy` and tell it to publish the `dist` directory, to manually deploy the build to a unique address that will be generated at the time of deploy.

guergana: I thought this can be inferred from the previous paragraph. But if it's clearer for you adding it as well to this paragraph I can add it.

lucaswerkmeister: I’d add it here and remove it from the previous paragraph to avoid duplication :) but I think it helps if it’s closer to the deploy command.

guergana: i think it's good to let the users know where the bundle is being built. this is why i thought it's a better idea to put it in the build step.

guergana: maybe we can ask a third person what they think? :)

README.md
```diff
@@ -11,3 +11,18 @@ _**Prerequisits: This repository requires node v16 and up**_
 1. Clone this repository
 2. Install dependecies with `npm i`
 3. Test and lint with `npm test`
+
+## Manual Deployment
```
itamargiv, line 15: This deploys the current commit to a uniquly generated URL, right? Maybe this is a more accurate title:

```suggestion
## Deploy Local Commits
```

README.md
itamargiv: I think I see the issue. I wonder if we should go by the old principle of "show, don't tell" here. What if we add a script that actually does all the steps for the developer, and then just instruct them to `npm run deployCommit` or something similar?

I think we can use both `buildPreview` and `netlify deploy --dir="./dist"` for that respect

README.md
```diff
@@ -11,3 +11,16 @@ _**Prerequisits: This repository requires node v16 and up**_
...
+## Deploy Local Commits
+
+### The first time you run a deploy
+
+To be able to run the netlify commands you will need to run `npm run netlify login` once after installing netlify to authenticate and obtain a token.
+
+### Every time when doing a deploy
+
+To create a deploy of the current commit, run `npm run deployCommit`. This will create a build in the `./dist` folder and create a manual deployment to a unique address that will be generated at the time of deploy. 
+To deploy to [https://new-lexeme-special-page.netlify.app](https://new-lexeme-special-page.netlify.app) run `npm run netlify deploy --prod`.
```
itamargiv, line 24: Since we don't have the `netlify` script anymore I think it would be:

```suggestion
To deploy to [https://new-lexeme-special-page.netlify.app](https://new-lexeme-special-page.netlify.app) run `npx netlify deploy --prod`.
```

README.md
```diff
@@ -11,3 +11,16 @@ _**Prerequisits: This repository requires node v16 and up**_
...
+
+## Deploy Local Commits
+
+### The first time you run a deploy
+
+To be able to run the netlify commands you will need to run `npm run netlify login` once after installing netlify to authenticate and obtain a token.
+
+### Every time when doing a deploy
+
+To create a deploy of the current commit, run `npm run deployCommit`. This will create a build in the `./dist` folder and create a manual deployment to a unique address that will be generated at the time of deploy.
```
itamargiv, line 23: Nice and snazzy! Thank you!

README.md
```diff
@@ -11,3 +11,16 @@ _**Prerequisits: This repository requires node v16 and up**_
 1. Clone this repository
 2. Install dependecies with `npm i`
 3. Test and lint with `npm test`
+
+## Deploy Local Commits
+
+### The first time you run a deploy
+
+To be able to run the netlify commands you will need to run `npm run netlify login` once after installing netlify to authenticate and obtain a token.
```
itamargiv, line 19: Same here, I think we could go with just `npx`:

```suggestion
To be able to run the netlify commands locally you will need to run `npx netlify login` once after `npm install` to authenticate and obtain a token.
```

itamargiv (commented):
Looking good, I have a couple of suggestions, but afterwards we are g2g :memo:

README.md
guergana: true!

README.md
```diff
@@ -11,3 +11,16 @@ _**Prerequisits: This repository requires node v16 and up**_
...
+## Deploy Local Commits
+
+### The first time you run a deploy
+
+To be able to run the netlify commands locally you will need to run `npx netlify login` once after `npm install` to authenticate and obtain a token.
+
+### Every time when doing a deploy
+
+To create a deploy of the current commit, run `npm run deployCommit`. This will create a build in the `./dist` folder and create a manual deployment to a unique address that will be generated at the time of deploy. 
+To deploy to [https://new-lexeme-special-page.netlify.app](https://new-lexeme-special-page.netlify.app) run `npx netlify deploy --prod`.
```
lucaswerkmeister, line 24: Shouldn’t this command also have `--dir='./dist'`? (Or maybe `npm run deployCommit --prod` would also work?)

guergana: also the current command works.

guergana: with the current state of the deployCommit script. `npm run deployCommit --prod` doesn't deploy to production.

guergana: it deploys by default to `./dist`. i dont understand the need of adding it to three places now.

lucaswerkmeister: That’s confusing, does it not ask about the directory in that case? (I haven’t tried it out yet.)

lucaswerkmeister: Here’s what I get:

```
$ npx netlify deploy --prod
Please provide a publish directory (e.g. "public" or "dist" or "."):
/var/www/html/wiki1/extensions/WikibaseLexeme/resources/special/new-lexeme
? Publish directory (.) 
```

lucaswerkmeister: Maybe it remembers your choice, and that’s why, on your end, it deploys `./dist` by default?

guergana: The directory is specified once when you login for the first time. You won't be asked for the directory again.

lucaswerkmeister: I do have an account, it made me create one. I don’t think it asked me about this setting at any point.

lucaswerkmeister: > with the current code of the deployCommit script. `npm run deployCommit --prod` doesn't deploy to production.

My bad, needs two more dashes. `npm run deployCommit -- --prod` seems to do the right thing (but I get an error, because my personal account isn’t allowed to deploy to this site, of course).

itamargiv: I would suggest removing this instruction altogether, since we will most probably not need to deploy to any production site. We only want commit deployments. Our "prod" is WikibaseLexeme itself.

lucaswerkmeister (approved):

PR #36, opened 2022-02-14
Lint for consistent component casing

lucaswerkmeister:
With kebab-case for now, by current majority opinion.


itamargiv (approved):
Thank you!

PR #39, opened 2022-02-15
Add basic input for lexical category

lucaswerkmeister:
Just a plain text field for now, the proper lookup will come later.

Also, add `await` in the integration test, my IDE tells me `setValue()` should be awaited (though the tests seem to work without it).

Bug: T301783

### DIFF
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -4,7 +4,11 @@ import { useStore } from 'vuex';
 import { Button as WikitButton } from '@wmde/wikit-vue-components';
 import { useMessages } from '@/plugins/MessagesPlugin/Messages';
 import LemmaInput from '@/components/LemmaInput.vue';
-import { UPDATE_LEMMA } from '@/store/actions';
+import LexicalCategoryInput from '@/components/LexicalCategoryInput.vue';
+import {
+	UPDATE_LEMMA,
+	UPDATE_LEXICAL_CATEGORY,
+} from '@/store/actions';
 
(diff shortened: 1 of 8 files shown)


src/components/NewLexemeForm.vue
```diff
@@ -4,7 +4,11 @@ import { useStore } from 'vuex';
 import { Button as WikitButton } from '@wmde/wikit-vue-components';
 import { useMessages } from '@/plugins/MessagesPlugin/Messages';
 import LemmaInput from '@/components/LemmaInput.vue';
-import { UPDATE_LEMMA } from '@/store/actions';
+import LexicalCategoryInput from '@/components/LexicalCategoryInput.vue';
+import {
+	UPDATE_LEMMA,
+	UPDATE_LEXICAL_CATEGORY,
```
itamargiv, line 10: I actually wonder if we truly need actions for these, are we planning these update procedures to be `async` in the near future? Otherwise, we can just cut the middleman and use the mutations directly (luckily this is not redux :D)

lucaswerkmeister: I’d also be okay with committing mutations directly, I think… does anyone else have objections to that?

micgro42 (approved):
Looks good! We can switch to mutations in a separate PR, to keep the focus of this one sharp.

src/components/NewLexemeForm.vue
lucaswerkmeister: Mutations done in #41.

PR #40, opened 2022-02-15
Add some margins to inputs

lucaswerkmeister:
And do that on an outer `<div>`, since Wikit resets the padding/marging of its own `<div>`s to 0.

### DIFF
--- a/src/components/LemmaInput.vue
+++ b/src/components/LemmaInput.vue
@@ -39,14 +39,23 @@ export default {
 </script>
 
 <template>
-	<text-input
-		class="mw-wbl-snl-lemma-input"
-		:label="messages.get( 'wikibaselexeme-newlexeme-lemma' )"
-		:placeholder="messages.get( 'wikibaselexeme-newlexeme-lemma-placeholder' )"
-		name="lemma"
-		required
-		:error="buildError( error )"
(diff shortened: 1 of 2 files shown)


src/components/LemmaInput.vue
```diff
@@ -39,14 +39,23 @@ export default {
...
-		class="mw-wbl-snl-lemma-input"
-		:label="messages.get( 'wikibaselexeme-newlexeme-lemma' )"
-		:placeholder="messages.get( 'wikibaselexeme-newlexeme-lemma-placeholder' )"
-		name="lemma"
-		required
-		:error="buildError( error )"
-		:value="modelValue"
-		@input="$emit( 'update:modelValue', $event )"
-	/>
+	<div class="mw-wbl-snl-lemma-input">
```
itamargiv, line 42: I like where you're coming from here, this enables us to be more flexible in case particular inputs have different spacing requirements.

But I'm not sure if the extra div is worth it. Especially since we could utilize flex and grid to manipulate the layout on the outer layer of the form, which also might be more appropriate in terms of separation of concerns.

lucaswerkmeister:
Replaced by #42 (which ended up with a very similar `margin-block-end` approach anyways).

PR #41, opened 2022-02-15
Directly use mutations in NewLexemeForm

lucaswerkmeister:
The actions aren’t worth the overhead here. Leave some commented-out guidance behind in `actions.ts`, because we’ll surely want some actions again at some point.

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -6,23 +6,27 @@
  * @see https://vuex.vuejs.org/guide/structure.html
  */
 
+/*
+ * Actions should be used for complex or asynchronous changes,
+ * otherwise it’s preferred to directly use mutations.
+ * Currently, there are no actions, but see the commented-out code below for guidance.
+ */
+
+/*
 import { ActionContext } from 'vuex';
 import RootState from './RootState';
 import {
-	SET_LEMMA,
-	SET_LEXICAL_CATEGORY,
+	MUTATION_NAME,
 } from './mutations';
+*/
 
-type RootContext = ActionContext<RootState, RootState>;
+// type RootContext = ActionContext<RootState, RootState>;
 
-export const UPDATE_LEMMA = 'updateLemma';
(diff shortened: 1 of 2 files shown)


src/store/actions.ts
```diff
@@ -6,23 +6,15 @@
...
+// export const ACTION_NAME = 'actionName';
 
 export default {
-	[ UPDATE_LEMMA ]( { commit }: RootContext, lemma: string ): void {
-		commit( SET_LEMMA, lemma );
-	},
-	[ UPDATE_LEXICAL_CATEGORY ]( { commit }: RootContext, lexicalCategory: string ): void {
-		commit( SET_LEXICAL_CATEGORY, lexicalCategory );
-	},
+	/* [ ACTION_NAME ]( { commit }: RootContext, payload ) { ... } */
```
lucaswerkmeister, line 19: Or should we leave this out and rely on the git history next time?

itamargiv: I guess we should fully rely on git and bring back the example we had here before, with the delayed increment

lucaswerkmeister: But that means also bringing back the counter in the root state, which doesn’t seem useful…

itamargiv: Good point. I guess just more context is warrented, that is, to make it clear that actions are to be used for async operations. and in conjunction with imported mutations.

lucaswerkmeister: Alright, I’ve added a comment in that direction and removed the commented-out code.

itamargiv: Apologies, I didn't make myself clear enough. I meant to add the context alongside the commented usage examples, this way in the diff of adding the first action, it would be more visible for comparison.

lucaswerkmeister: I think you were clear, I just decided for myself that the commented-out code was redundant with the Git history :P but your point about code review when adding the next action is valid, so I can add it back in.

itamargiv (approved):
Thank you!

PR #42, opened 2022-02-15
Add spacing between the components to the form

micgro42:
The button needs a wrapping `<div>` so it doesn't get stretched to the
full width. (This button really should also be its own component).


src/components/NewLexemeForm.vue
```diff
@@ -46,20 +46,26 @@ export default {
...
 			v-model="lexicalCategory"
 		/>
-		<wikit-button
-			class="form-button-submit"
-			type="progressive"
-			variant="primary"
-		>
-			{{ $messages.get( 'wikibaselexeme-newlexeme-submit' ) }}
-		</wikit-button>
+		<div>
```
itamargiv, line 49: I think we should consider this the footer of the form, or at least a section of it.

src/components/NewLexemeForm.vue
```diff
@@ -46,20 +46,26 @@ export default {
...
 	</form>
 </template>
 
 <style scoped lang="scss">
 @import "@wmde/wikit-tokens/variables";
 
 .new-lexeme-form {
+	// display
+	display: grid;
+	gap: $dimension-layout-xsmall;
```
itamargiv, line 67: Wonderful! :rainbow:

src/components/NewLexemeForm.vue
micgro42: Sure. Is there anything you want me to change for that?

itamargiv: Sorry, I didn't make the actionable clear. Yeas, I think we should use an HTML `<footer>` element.

micgro42: Mh, not sure if I would go that far, to be honest. This is still a central component of the form that just happens to appear at the form's end.

A `<footer>` seems to usually contain a different class of content. From the [spec](https://html.spec.whatwg.org/multipage/sections.html#the-footer-element):

> A footer typically contains information about its section such as who wrote it, links to related documents, copyright data, and the like.

The submit button of a form doesn't feel like it belongs into that list.

src/components/NewLexemeForm.vue
```diff
@@ -46,20 +46,26 @@ export default {
...
+		</div>
 	</form>
 </template>
 
 <style scoped lang="scss">
 @import "@wmde/wikit-tokens/variables";
 
 .wbl-snl-form {
+	// display
+	display: grid;
```
lucaswerkmeister, line 66: Are you sure we can use grid? MDN says it’s supported since Safari 10.1, and [browser support matrix Grade A](https://www.mediawiki.org/wiki/Compatibility#Browser_support_matrix) (and [browserslist-config-wikimedia/modern](https://github.com/wikimedia/browserslist-config-wikimedia/blob/v0.3.0/modern.js#L9)) still include Safari 9.1 :/

micgro42: Our reference is [modern-es6-only](https://github.com/wikimedia/browserslist-config-wikimedia/blob/v0.3.0/modern-es6-only.js) as only that covers browsers that support the Proxy object and thus Vue 3. But that still includes Safari 10 and `display: grid` is apparently only supported in Safari 10.1 :persevere: 

I'll change it to flex-box :disappointed: 

/me should have been bold when creating that `modern-es6-only.js` file and trying to sneak in a _slightly_ newer Safari version :/

micgro42: updated. Please have another look

src/components/NewLexemeForm.vue
```diff
@@ -46,20 +46,27 @@ export default {
...
 </template>
 
 <style scoped lang="scss">
 @import "@wmde/wikit-tokens/variables";
 
 .wbl-snl-form {
+	// display
+	display: flex;
+	flex-direction: column;
+	gap: $dimension-layout-xsmall;
```
lucaswerkmeister, line 68: Hm, [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/gap#browser_compatibility) claims that `gap` support in Flex Layout is actually really recent? Chrome 84, Firefox 63, Safari 14.1…

Maybe there’s an older, flex-specific property we should use instead?

src/components/NewLexemeForm.vue
micgro42: Also ... I'm not sure vite actually makes use of `.browserlistrc`?

src/components/NewLexemeForm.vue
micgro42: :sob:

micgro42: Now it should work everywhere, ugly as it is

src/components/NewLexemeForm.vue
lucaswerkmeister: I proposed a Codex change, it turns out to affect stylelint at least: https://gerrit.wikimedia.org/r/c/design/codex/+/763199

src/components/NewLexemeForm.vue
```diff
@@ -46,20 +46,26 @@ export default {
...
+			</wikit-button>
+		</div>
 	</form>
 </template>
 
 <style scoped lang="scss">
 @import "@wmde/wikit-tokens/variables";
 
 .wbl-snl-form {
+	& > *:not(:last-child) {
```
lucaswerkmeister, line 65: The `*` here is redundant, right? (It also seems to be compiled away anyways.)

lucaswerkmeister (dismissed):
This works because `.wbl-snl-form > :not(:last-child)` has higher specificity than Wikit’s `.wikit, .wikit * { margin: 0; }`, whereas in #40 I had to use a wrapping div. I’m not sure if I like adding this margin directly to divs that Wikit doesn’t want to have any margin – maybe we should add wrapper divs of our own anyways, and only apply the margin to those? But that can also be done later.

src/components/NewLexemeForm.vue
micgro42: Apparently it is :astonished: . Though it is not being compiled away for me. The CSS in the browser changes for me when the `*` is removed.

lucaswerkmeister (dismissed):

src/components/NewLexemeForm.vue
micgro42: Wonderful indeed. Sadly, it didn't last 😔

src/components/NewLexemeForm.vue
```diff
@@ -46,20 +46,26 @@ export default {
...
 	</form>
 </template>
 
 <style scoped lang="scss">
 @import "@wmde/wikit-tokens/variables";
 
 .wbl-snl-form {
+	& > :not(:last-child) {
+		margin-block-end: $dimension-layout-xsmall;
+	}
```
itamargiv, line 67: Personally, to recreate this grid functionality without getting into specificity loopholes of `:not`, we used to use the "owl eyes"

```suggestion
	& > * + * {
		margin-block-start: $dimension-layout-xsmall;
	}
```

src/components/NewLexemeForm.vue
itamargiv: Sure, just a suggestion :)

src/components/NewLexemeForm.vue
micgro42: looks unusual, but why not. I'll update the PR

itamargiv (commented):
Looks great, works! Sorry for the long review, should've picked up on our browser support matrix before.

itamargiv (approved):

src/components/NewLexemeForm.vue
lucaswerkmeister: FWIW, to me this looks much less readable than the old version :/

PR #44, opened 2022-02-16
Deploy to Netlify for every pull request and main

micgro42:
* [ ] This still needs some setup of env variables in GitHub

Bug: T301801


micgro42:
replaced by #45

PR #45, opened 2022-02-16
Deploy to Netlify for every pull request and main

micgro42:
New pull request to test the modified `Comment deploy URL to PR` job

Bug: T301801


.github/workflows/demo.yaml
```diff
@@ -0,0 +1,36 @@
...
+              uses: actions/setup-node@v2
+              with:
+                  node-version: "16"
+            - name: Install dependencies
+              run: npm ci
+            - name: Build Project as app
+              run: npm run buildPreview
+            - name: Deploy to netlify
+              id: netlify_deploy
+              run: npx -p netlify-cli
```
lucaswerkmeister, line 31: Do we need the `-p netlify-cli` when it should already be installed as part of the `npm ci` earlier?

.github/workflows/comment-deploy-url-to-pr.yaml
```diff
@@ -0,0 +1,19 @@
...
+        types: [opened]
+
+jobs:
+    comment-on-pr:
+        runs-on: ubuntu-latest
+        name: Comment deploy URL to PR
+        steps:
+            - name: Get branch hash
+              id: get_hash
+              run: echo "::set-output name=hash::$(echo '${{ github.ref }}' | shasum | awk '{print substr($1,0,36)}')"
```
lucaswerkmeister, line 14: I think some comment here, pointing to the other workflow where the same hash is generated, would be useful. (And maybe in the other direction as well, instead of “a staging area”.)

lucaswerkmeister (commented):
looks pretty good, minor comments

.github/workflows/demo.yaml
micgro42: Good point. I wonder if could go in the opposite direction though? If we add `-p netlify-cli` also to the other command, maybe we can drop the dev dependency on Netlify and thus get rid of those vulnerability notices. That would seem to be a good trade-off given that most of use will likely not use the manual deploy much, since we don't even have the needed credentials.

.github/workflows/comment-deploy-url-to-pr.yaml
micgro42: Done.

.github/workflows/demo.yaml
lucaswerkmeister: I’d be quite happy with that, it should also make `npm install` much faster. @guergana WDYT?

guergana: If we dont need to install it locally then we shouldn't do it. go for it! :)

lucaswerkmeister (approved):

PR #47, opened 2022-02-17
Add licensing information

chukarave:
While working on this it became apparent we are going to have to get some of the information from the RepoSettings in order to display the CC0 license link correctly. 
This PR reflects the new-lexeme-special-page side of receiving the parameters, initiating the store with them and using them within the component.

Bug: [T298152](https://phabricator.wikimedia.org/T298152)

### DIFF
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -86,6 +86,7 @@ export default {
 		/>
 <<<<<<< HEAD
 <<<<<<< HEAD
+<<<<<<< HEAD
 		<input
 			type="hidden"
 			name="wpEditToken"
@@ -97,6 +98,9 @@ export default {
 =======
 		<p class="wbl-snl-copyright wbl-snl-small-text" v-html="copyrightText" />
 >>>>>>> Add styles to the copyright info
+=======
+		<p class="wbl-snl-copyright" v-html="copyrightText" />
+>>>>>>> removing small text class and unifying mixin inclusion under copyright class
 		<div>
 			<wikit-button
 				class="form-button-submit"
@@ -129,15 +133,21 @@ export default {
 }
 
 <<<<<<< HEAD
+<<<<<<< HEAD
 .wbl-snl-copyright {
 	@include small-text;
 =======
 .wbl-snl-small-text {
+=======
+.wbl-snl-copyright {
+>>>>>>> removing small text class and unifying mixin inclusion under copyright class
 	@include small-text;
-}
 
+<<<<<<< HEAD
 .wbl-snl-copyright {
 >>>>>>> Add styles to the copyright info
+=======
+>>>>>>> removing small text class and unifying mixin inclusion under copyright class
 	font-style: italic;
 }
 
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -40,6 +40,7 @@ const lexicalCategory = computed( {
 } );
 const token = computed( () => store.state.token );
 const submitMsg = $messages.get( 'wikibaselexeme-newlexeme-submit' );
+<<<<<<< HEAD
 const termsOfUseTitle = $messages.get( 'copyrightpage' );
 const copyrightText = $messages.get(
 	'wikibase-shortcopyrightwarning',
@@ -47,6 +48,15 @@ const copyrightText = $messages.get(
 	termsOfUseTitle,
 	store.state.config.licenseUrl,
 	store.state.config.licenseName,
+=======
+const termsOfUseURL = 'https://foundation.wikimedia.org/wiki/Terms_of_Use';
+const copyrightText = $messages.get(
+	'wikibase-shortcopyrightwarning',
+	submitMsg,
+	termsOfUseURL,
+	store.state.config.cc0Url,
+	store.state.config.cc0Text,
+>>>>>>> Add cc0 link text and url to store and inject into createAndMount
 );
 </script>
 
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -108,6 +108,7 @@ export default {
 
 .wbl-snl-copyright {
 	@include small-text;
+
 	font-style: italic;
 }
 
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -40,7 +40,6 @@ const lexicalCategory = computed( {
 } );
 const token = computed( () => store.state.token );
 const submitMsg = $messages.get( 'wikibaselexeme-newlexeme-submit' );
-<<<<<<< HEAD
 const termsOfUseTitle = $messages.get( 'copyrightpage' );
 const copyrightText = $messages.get(
 	'wikibase-shortcopyrightwarning',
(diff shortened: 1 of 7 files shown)


src/components/NewLexemeForm.vue
```diff
@@ -28,6 +28,16 @@ const lexicalCategory = computed( {
 		store.commit( SET_LEXICAL_CATEGORY, newLexicalCategory );
 	},
 } );
+
+const submitMsg = $messages.get( 'wikibaselexeme-newlexeme-submit' );
+const termsOfUseURL = 'https://foundation.wikimedia.org/wiki/Terms_of_Use';
```
lucaswerkmeister, line 33: Hard-coding this doesn’t seem right to me – in Data Bridge, we get it from the custom `wbdatabridgeconfig` API query+meta module, where it comes from parsing the [copyrightpage message](https://www.wikidata.org/wiki/MediaWiki:Copyrightpage) as a title.

src/components/NewLexemeForm.vue
```diff
@@ -46,15 +56,14 @@ export default {
...
-			<wikit-button
-				class="form-button-submit"
-				type="progressive"
-				variant="primary"
-			>
-				{{ $messages.get( 'wikibaselexeme-newlexeme-submit' ) }}
-			</wikit-button>
-		</div>
+		<p class="wbl-snl-copyright" v-html="copyrightText" />
+		<wikit-button
```
lucaswerkmeister, line 60: Is it intentional that the button is no longer wrapped in a `<div>`?

src/components/NewLexemeForm.vue
```diff
@@ -28,6 +28,16 @@ const lexicalCategory = computed( {
...
+
+const submitMsg = $messages.get( 'wikibaselexeme-newlexeme-submit' );
+const termsOfUseURL = 'https://foundation.wikimedia.org/wiki/Terms_of_Use';
+const copyrightText = $messages.get(
+	'wikibase-shortcopyrightwarning',
+	submitMsg,
+	termsOfUseURL,
+	store.state.config.licenseUrl,
+	store.state.config.licenseName,
+);
```
lucaswerkmeister, line 40: This doesn’t seem to match the version of the `wikibase-shortcopyrightwarning` message that I see in `repo/i18n/en.json` – that message only has three parameters, not four, and the second one looks like it should be a title instead of a full URL.

chukarave:
Putting this back to draft, as this change requires changes from multiple repositories before it can merged

src/components/NewLexemeForm.vue
```diff
@@ -88,4 +100,13 @@ export default {
...
 	border-color: $border-color-base-subtle;
 }
+
+.wbl-snl-small-text {
+	@include small-text;
+}
+
+.wbl-snl-copyright {
+	font-style: italic;
+}
```
lucaswerkmeister, line 110: I assume we could also merge this, and use just the one class:

```suggestion
.wbl-snl-copyright {
	@include small-text;
	font-style: italic;
}
```

To me that would seem slightly nicer, but the two classes are also okay. WDYT?

src/store/index.ts
```diff
@@ -4,20 +4,28 @@
...
-			language: '',
-			lexicalCategory: '',
-		};
-	},
-	mutations,
-	actions,
-} );
+type StoreParams = { licenseUrl?: string, licenseName?: string };
+
+export default function initStore( { licenseUrl = '', licenseName = '' }: StoreParams ): Store<RootState> {
```
lucaswerkmeister, line 15: Editorial note: this `initStore()` refactoring was extracted into #51 and already merged.

chukarave: Hmm. But it doesn't include the license text and url yet does it?

src/components/NewLexemeForm.vue
chukarave: There's no reason why it shouldn't all be in wbl-snl-copyright, but Sarai suggested a small-text class for the mixin so I followed that.

sai-san: Hey. I suggested creating separate text style classes so a style can be applied to multiple elements without having to include the same mixin every time. This won't be the case for  the `small-text` style, but `body` will for sure be reused. I thought this approach sounded more efficient, but it's just an inexperienced assumption.

chukarave: I think your suggestion is definitely efficient for a bigger component with more text, this one only has the license text in it though, if I'm not mistaking.

src/store/index.ts
lucaswerkmeister: Nope, just the change from exporting a store instance to a function.

src/components/NewLexemeForm.vue
sai-san: We'll have to apply quite a few styles! Not all the styles will be used repeatedly, but still it would be better to take a unified approach.

Screenshot 2022-02-25 at 10 34 30
s.

chukarave: thanks @sai-san!

src/components/NewLexemeForm.vue
chukarave: see latest commit

src/components/NewLexemeForm.vue
```diff
@@ -38,6 +38,16 @@ const lexicalCategory = computed( {
 		store.commit( SET_LEXICAL_CATEGORY, newLexicalCategory );
 	},
 } );
+
+const submitMsg = $messages.get( 'wikibaselexeme-newlexeme-submit' );
+const termsOfUseURL = $messages.get( 'copyrightpage' );
```
lucaswerkmeister, line 43: I suggest renaming this (e.g. to `termsOfUseURL`), since it won’t actually be a URL anymore.

src/components/NewLexemeForm.vue
```diff
@@ -39,6 +39,15 @@ const lexicalCategory = computed( {
...
 } );
 const token = computed( () => store.state.token );
+const submitMsg = $messages.get( 'wikibaselexeme-newlexeme-submit' );
+const termsOfUseTitle = $messages.get( 'copyrightpage' );
+const copyrightText = $messages.get(
+	'wikibase-shortcopyrightwarning',
+	submitMsg,
+	termsOfUseTitle,
+	store.state.config.licenseUrl,
+	store.state.config.licenseName,
```
lucaswerkmeister, line 49: I wonder if we should make this message `computed()`, because the store state could theoretically change? (We don’t expect it to actually change, but the same is true of the token, immediately above, and that is still computed.)

src/store/index.ts
```diff
@@ -14,11 +14,15 @@ import mutations from './mutations';
 import RootState from './RootState';
 
 interface StoreParams {
-	token?: string;
+	token?: string,
+	licenseUrl?: string,
+	licenseName?: string;
```
lucaswerkmeister, line 19: We have two lines ending with `,` and one ending with `;` here now, that seems inconsistent to me. (Would be nice to have an eslint rule for it, to fix it automatically.)

src/components/NewLexemeForm.vue
lucaswerkmeister: Looks like the rule is [member-delimiter-style](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/member-delimiter-style.md), and it’s not part of our config at the moment. I’ll open a separate PR to propose adding it.

lucaswerkmeister: wmde/eslint-config-wikimedia-typescript#15

src/store/index.ts
lucaswerkmeister: Now in #77.

src/components/NewLexemeForm.vue
itamargiv: I don't think this would ever change, we are inheriting this from outside the app, and only using the store to avoid drilling down properties at this point, but this raises a good point that I think we discussed a few weeks ago which is to create a specialized plugin just for the config something that we decided to do down the line. Please feel free to create a task for that on the backlog, though

lucaswerkmeister: Created T303050. Let’s leave this as it is in the meantime.

lucaswerkmeister (approved):

PR #48, opened 2022-02-17
Layout for the form in context of the special page

micgro42:
This creates the basic layout of the special page in context of creating the basic form for T301234.

The message component is currently just a placeholder. When work on that message properly, we'll extract it into a separate component.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,10 +1,52 @@
 <script setup lang="ts">
-// This starter template is using Vue 3 <script setup> SFCs
-// Check out https://v3.vuejs.org/api/sfc-script-setup.html#sfc-script-setup
 import NewLexemeForm from '@/components/NewLexemeForm.vue';
+import { Message as WikitMessage } from '@wmde/wikit-vue-components';
 import '@wmde/wikit-vue-components/dist/wikit-vue-components.css';
 </script>
 
 <template>
-	<new-lexeme-form />
+	<p class="wbl-snl-intro-text">
+		To Be Done
+	</p>
+	<div class="wbl-snl-app">
+		<new-lexeme-form />
+		<wikit-message type="notice">
+			To Be Done
+		</wikit-message>
+	</div>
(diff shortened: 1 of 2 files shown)


micgro42:
Preview at: https://3f5163aa9805d4967e7062d18f95c56667d2--new-lexeme-special-page.netlify.app

vite.config.ts
```diff
@@ -36,6 +36,7 @@ export default defineConfig( {
 				__dirname,
 				'node_modules/@wmde/wikit-vue-components/dist/wikit-vue-components.css',
 			),
+			'@wmde/wikit-vue-components/src/styles/mixins/Typography': '@wmde/wikit-vue-components/src/styles/mixins/Typography',
```
lucaswerkmeister, line 39: Hm, does that mean the previous alias also doesn’t need the `resolve()` call?

src/App.vue
```diff
@@ -1,10 +1,52 @@
...
+
+	@media (min-width: $break-point) {
+		align-items: stretch;
+
+		& > * + * {
+			margin-inline-start: $dimension-layout-medium;
+		}
+	}
+
+	@media (max-width: $break-point) {
```
lucaswerkmeister, line 40: Doesn’t one of these two `@media` queries need to be offset by one? Otherwise at exactly 981px width, both will apply.

vite.config.ts
micgro42: Yeah, maybe :thinking:

src/App.vue
micgro42: Good point, Figma says `981px and below`, so the other one get's a `+1`

micgro42: Now it should work

lucaswerkmeister (approved):

vite.config.ts
lucaswerkmeister: #49

PR #50, opened 2022-02-17
POC: Submit to Special:NewLexeme

lucaswerkmeister:
This makes the form functional (in a way), by submitting to the backend
of the old special page, Special:NewLexeme – we were careful to use the
same HTML input names for the lemma, lexeme language, and lexical
category, so with a hidden input for the lemma language (hard-coded to
en for now) and another one for the CSRF token, this actually works.

We probably don’t want to actually go this route even as a basic first
version on the main branch, but for experimenting it’s useful.

Bug: T298153


lucaswerkmeister:
Superseded by #55.

PR #52, opened 2022-02-18
Add hidden token input to form

lucaswerkmeister:
Using a token passed into `createAndMount()`.

Bug: T302001

---

This doesn’t include a test for `createAndMount` passing the token into the store, because I’m not sure how to test that. Should we test (again) that there’s an input with the right value? (But then `main.test.ts` probably belongs in `integration/` rather than `unit/`.)


itamargiv (approved):
LGTM thnak you!

PR #53, opened 2022-02-18
Add mocks for all messages for the dev entry point

micgro42:


src/plugins/MessagesPlugin/MockMessagesRepository.ts
```diff
@@ -0,0 +1,24 @@
...
+	'wikibaselexeme-newlexeme-lexicalcategory-placeholder': 'The Lexeme\'s category, e.g. \'numeral\'',
+	'wikibaselexeme-newlexeme-submit': 'Create Lexeme',
+	'wikibaselexeme-newlexeme-error-no-lemma': 'Lemma field is empty, please make an entry.',
+	'wikibaselexeme-newlexeme-error-lemma-is-too-long': 'FIXME (copy is missing!)',
+};
+
+export class MockMessagesRepository implements MessagesRepository {
+
+	public get( key: MessageKeys ): string {
+		return messages[ key ] !== undefined ? messages[ key ] : `⧼${key}⧽`;
```
lucaswerkmeister, line 19: Maybe this one should HTML-escape the return value, so we don’t rely on the hard-coded fallbacks being safe? (Could also be done separately, though.)

lucaswerkmeister (approved):

src/plugins/MessagesPlugin/MockMessagesRepository.ts
micgro42: I looked into this again for a bit, and I think I'd like to do the escaping the text in an extra PR. Do you have a favorite way to escape strings in JS? Somehow, I'm not yet fully convinced by the methods that I've seen so far.

lucaswerkmeister: MediaWiki’s `mw.html.escape()` replaces `'`, `"`, `<`, `>`, `&` with their entities; `Html::element()` even only replaces `&` and `<`, since everything else doesn’t need to be escaped in element contents. I think the latter might also be enough for us, since we’ll probably only put messages in element contents, using `v-html`? (I’m not even sure how we would add a message in an attribute without it being escaped by Vue.)

micgro42: > MediaWiki’s `mw.html.escape()`

Mh, I guess we can copy and paste that method into this class and use it here :thinking:

PR #54, opened 2022-02-21
Replace leftover mw-snl-* classes with just snl-*

lucaswerkmeister:
Compare #43 / 5b8a5cd91c.


itamargiv (approved):
Interestingly enough, this isn't caught be the linter, since no style rules were declared for these classes. This brings up the question of whether it would also be advisable to create a followup implementing some way to check the classes on the DOM nodes of the vue component's template.

Either or, thanks for the cleanup

PR #55, opened 2022-02-21
Add lemma language input and submit form

lucaswerkmeister:
This is a temporary solution to make creating lexemes work, by targeting the noscript backend of the special page; we want to replace this with an API request later. (At some point the hard-coded lemma-language will also be replaced, of course.)

The previous simple Cypress test is replaced with one that watches for a form submit and checks the submitted data (including a token that we hard-code in index.html). Unfortunately, Chromium reports inexplicable cross-origin errors when submitting, blocking the iframe with origin http://localhost:3000 from accessing a cross-origin frame at http://localhost:3000, or something like that (yes, that’s the same host twice, as far as I can tell). I haven’t been able to find a solution for that, so for now, work around it by disabling Chrome web security in cypress.json. We can remove that line once we no longer submit the form.

Bug: T298153

---

See https://gerrit.wikimedia.org/r/c/mediawiki/extensions/WikibaseLexeme/+/763719 for the WikibaseLexeme part.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,9 @@
 		"url": "https://phabricator.wikimedia.org/project/board/5674/"
 	},
 	"homepage": "https://github.com/wmde/new-lexeme-special-page#readme",
+	"engines": {
+		"node": ">=16"
+	},
 	"dependencies": {
 		"@vue/compat": "^3.2.30",
 		"@wmde/wikit-tokens": "^2.1.0-alpha.12",
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,5 @@
 {
 	"baseUrl": "http://localhost:3000/",
+	"chromeWebSecurity": false,
 	"supportFile": false
 }
--- a/cypress/integration/NewLexemeForm.spec.js
+++ b/cypress/integration/NewLexemeForm.spec.js
@@ -1,11 +1,32 @@
 describe( 'NewLexemeForm', () => {
 
-	it( 'adds a hidden input with the CSRF token', () => {
+	it( 'submits form data', () => {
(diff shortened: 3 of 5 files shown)


itamargiv (dismissed):
So far looks great. I suppose at this point, we should probably start devising some cypress browser tests, to check that it all works. We can do this by triggering the button click (or submitting the form directly) and intercepting the URL it submits to.

See: https://docs.cypress.io/api/commands/intercept
See: https://docs.cypress.io/api/commands/submit

lucaswerkmeister:
> Done, see three new commits. (Could maybe be split out into separate PRs too.)

Cypress setup split out into #56. (Single PR with there commits – I’m hoping it’s okay to merge it without squashing.)

package.json
```diff
@@ -40,6 +40,9 @@
 		"url": "https://phabricator.wikimedia.org/project/board/5674/"
 	},
 	"homepage": "https://github.com/wmde/new-lexeme-special-page#readme",
+	"engines": {
+		"node": ">=16"
+	},
```
lucaswerkmeister, line 45: I can’t fully remember why this is here… I think at some point I got a warning about compatibility with Node 8 (which I guess is the default version if the `package.json` doesn’t specify engines?), but I can’t reproduce this now.

micgro42: Doesn't hurt, and it is consistent with our `.nvmrc`.

micgro42 (approved):

lucaswerkmeister:
Rebased.

micgro42 (approved):
I think this has only be rebased since my last approval without major changes, so I'm adding the Approve again.

PR #56, opened 2022-02-28
Set up Cypress tests for app

lucaswerkmeister:
See the individual commit messages.


cypress/plugins/index.js
```diff
@@ -7,6 +7,7 @@
  * @type {Cypress.PluginConfig}
  */
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 module.exports = ( on, config ) => {
```
micgro42, line 11: As I understand it, that typescript-eslint rule has the same option as the vanilla eslint rule: https://eslint.org/docs/rules/no-unused-vars

So unless we configured it away somewhere, it should be enough to change the parameters names to start with underscore for that rule to not trigger.

package.json
```diff
@@ -10,12 +10,12 @@
 		"buildPreview": "BUILD_AS_APP=1 vite build",
 		"fix": "run-s fix:*",
 		"fix:prettier": "prettier --write '**/*.{json,yml,yaml}'",
-		"fix:eslint": "eslint --fix --ext .js,.ts,.vue --ignore-path .gitignore . --ignore-pattern 'cypress'",
+		"fix:eslint": "eslint --fix --ext .js,.ts,.vue --ignore-path .gitignore .",
 		"fix:stylelint": "stylelint --fix '**/*.{css,scss,vue}'",
 		"deployCommit": "npm run buildPreview && npx -p netlify-cli netlify deploy --dir='./dist'",
 		"lint": "run-s lint:*",
 		"lint:prettier": "prettier --check '**/*.{json,yml,yaml}'",
-		"lint:eslint": "eslint --ext .js,.ts,.vue --ignore-path .gitignore . --ignore-pattern 'cypress'",
```
micgro42, line 18: Good, removing this ignore-pattern was on my mental todo-list as well. :+1:

cypress/plugins/index.js
lucaswerkmeister: Good point, done.

micgro42 (approved):

PR #75, opened 2022-03-01
Remove vue-tsc from build command

lucaswerkmeister:
This command doesn’t actually contribute to the build results, it just takes a lot of time (much longer than `vite build` itself). We already run this as part of the test command, so I don’t think there’s a reason to keep it in the build command as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"main": "main.ts",
 	"scripts": {
 		"dev": "vite",
-		"build": "vue-tsc --noEmit && vite build",
+		"build": "vite build",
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Part of the motivation for this is T302794: we want to run builds of this repository in WikibaseLexeme’s CI, but that still uses Node 12 (though that might change with T267890), and parts of vue-tsc require Node 14 (e.g. vscode-jsonrpc uses `?.` syntax). I didn’t mention this in the commit message because it’s possible that this isn’t enough, and other parts of the build will still require a newer Node.

micgro42 (approved):

PR #80, opened 2022-03-10
Add init() entry point with MwMessagesRepository

lucaswerkmeister:
WikibaseLexeme.git will no longer directly call `createAndMount()` (though that’s still used directly by the dev entry point and the Cypress tests), but instead the new `init()` function. This new entry point creates the MediaWiki-based `MessagesRepository` implementation, and will also create other MediaWiki-based service implementations later (whereas dev and tests might use different implementations).

Bug: T302961

### DIFF
--- a/src/@types/mediawiki.ts
+++ b/src/@types/mediawiki.ts
@@ -1,0 +1,9 @@
+export interface MwMessage {
+	text(): string;
+	parse(): string;
+}
+export type MwMessages = ( key: string, ...params: readonly ( string|HTMLElement )[] ) => MwMessage;
+
+export interface MediaWiki {
+	message: MwMessages;
+}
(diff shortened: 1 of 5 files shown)


src/@types/mediawiki.ts
```diff
@@ -0,0 +1,9 @@
+export interface MwMessage {
```
micgro42, line 1: Maybe it is still a bit too early, but I wonder if at some point we should upstream all these types to https://github.com/wikimedia/typescript-types/blob/master/MediaWiki.d.ts

micgro42 (approved):

src/@types/mediawiki.ts
lucaswerkmeister: I created https://github.com/wikimedia/typescript-types/pull/19 during a meeting earlier today, let’s see if anything happens there.

PR #86, opened 2022-03-16
Remove hidden token input from NewLexemeForm

lucaswerkmeister:
No longer needed now that we’re using the API to create lexemes, `postWithEditToken()` adds the token automatically.

Bug: T302961

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,7 +14,6 @@
 
 			const config = {
 				rootSelector: '#app',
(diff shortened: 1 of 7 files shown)


lucaswerkmeister:
Follow-up removing the token on the WikibaseLexeme side: https://gerrit.wikimedia.org/r/c/mediawiki/extensions/WikibaseLexeme/+/771380

PR #88, opened 2022-03-18
Move files out of mediawiki/ directory

lucaswerkmeister:
This distinction isn’t very helpful at the moment – it seems better to have all the data access stuff in `data-access` (and `MwMessagesRepository` in `MessagesPlugin/`).


chukarave (approved):
ok lgtm

PR #78, opened 2022-03-04
WIP: Lookups

micgro42:
Still very wip, but I noticed that somehow the Wikit keyboard navigation is somewhat broken:

1.: the options menu does not resize
2.: it doesn't find the highlighted element when trying to scroll it into view in https://github.com/wmde/wikit/blob/master/vue-components/src/components/OptionsMenu.vue#L142

PR #90, opened 2022-03-18
Prepare for searching items

lucaswerkmeister:
This can be used as a basis for the lookups (#78).

----

TODO:
- [x] tests for MwApiItemSearcher, at least

### DIFF
--- a/src/data-access/DevItemSearcher.ts
+++ b/src/data-access/DevItemSearcher.ts
@@ -1,0 +1,48 @@
+import {
+	commonParams,
+	processResponse,
+	WbSearchEntitiesResponse,
+} from '@/data-access/apiBasedItemSearcher';
+import ItemSearcher, { SearchedItemOption } from '@/data-access/ItemSearcher';
+
+/**
+ * Item searcher for the dev entry point.
+ * Searches Wikidata (hard-coded), and also supports some bespoke syntax:
+ * =Q123 will always return a single fake search result for that item ID.
+ * (This is meant to be used in the Cypress tests.
+ * A syntax for synthetic errors might also be added later.)
+ */
+export default class DevItemSearcher implements ItemSearcher {
+
+	public async searchItems( searchTerm: string, offset = 0 ): Promise<SearchedItemOption[]> {
+		const match = searchTerm.match( /^=(Q[1-9][0-9]*)$/ );
+		if ( match ) {
+			const itemId = match[ 1 ];
+			return [ {
+				itemId: itemId,
+				display: {
+					label: {
+						language: 'en',
+						value: `fake label for ${itemId}`,
+					},
+					description: {
+						language: 'en',
+						value: `fake description for ${itemId}`,
+					},
+				},
+			} ];
+		}
+
+		const response = await fetch( 'https://www.wikidata.org/w/api.php?' + new URLSearchParams( {
+			...commonParams,
+			search: searchTerm,
+			language: 'en',
+			continue: offset.toString(),
+			format: 'json',
+			origin: '*',
+		} ) ).then( ( r ) => r.json() ) as WbSearchEntitiesResponse;
+
+		return processResponse( response, 'en' );
+	}
+
+}
--- a/src/data-access/ItemSearcher.ts
+++ b/src/data-access/ItemSearcher.ts
@@ -1,0 +1,19 @@
+export interface SearchedItemOption {
+	itemId: string;
+	display: {
+		label?: {
+			language: string;
+			value: string;
+		}; // Term
+		description?: {
+			language: string;
+			value: string;
+		}; // Term
+	};
+}
+
+export default interface ItemSearcher {
+
+	searchItems( searchTerm: string, offset?: number ): Promise<SearchedItemOption[]>;
+
+}
--- a/src/data-access/MwApiItemSearcher.ts
+++ b/src/data-access/MwApiItemSearcher.ts
@@ -1,0 +1,31 @@
+import { MwApi } from '@/@types/mediawiki';
+import {
+	commonParams,
+	processResponse,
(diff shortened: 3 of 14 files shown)


src/data-access/MwApiItemSearcher.ts
```diff
@@ -0,0 +1,31 @@
...
+		this.api = api;
+		this.languageCode = languageCode;
+	}
+
+	public async searchItems( searchTerm: string, offset?: number ): Promise<SearchedItemOption[]> {
+		// TODO handle errors
+		const response = await this.api.get( {
+			...commonParams,
+			search: searchTerm,
+			language: this.languageCode,
```
micgro42, line 24: I _think_ this also requires a `uselang=this.languageCode` parameter. The language parameter only defines in which language the _search_ happens, and the _uselang_ parameter defines the language of the labels and descriptions and such that are being returned.

Or can we rely on cookies or something similar here?

See also the discussion in T109584

micgro42: (That being said, I suspect the `language` parameter being broken and effectively ignored right now. That's why it is searching Item labels and aliases in all languages.)

src/data-access/apiBasedItemSearcher.ts
```diff
@@ -0,0 +1,47 @@
+import { SearchedItemOption } from '@/data-access/ItemSearcher';
+
+/*
+ * Utility code shared by several API-based ItemSearcher implementations.
+ */
+
+export const commonParams: Record<string, string> = {
```
micgro42, line 7: Not sure if these are worth it, beyond the magic number. But not too harmful either, I guess.

src/data-access/MwApiItemSearcher.ts
lucaswerkmeister: You’re right, we definitely want `uselang=` for error messages at least (that applies to the LexemeCreator too). Maybe we can add that to the whole api instance as a default parameter? I’ll check.

src/data-access/apiBasedItemSearcher.ts
lucaswerkmeister: Yeah, it had felt like more common parameters when I was copying them :D I wouldn’t mind losing these either.

src/data-access/MwApiItemSearcher.ts
lucaswerkmeister: Merged and rebased :)

micgro42: cool, I'll have another look!

tests/unit/data-access/MwApiItemSearcher.test.ts
```diff
@@ -0,0 +1,95 @@
...
+import { SearchedItemOption } from '@/data-access/ItemSearcher';
+import MwApiItemSearcher from '@/data-access/MwApiItemSearcher';
+import unusedApi from '../../mocks/unusedApi';
+
+describe( 'MwApiItemSearcher', () => {
+
+	it( 'gets and transforms search results', async () => {
+		const api: MwApi = {
+			...unusedApi,
+			get: jest.fn().mockResolvedValue( {
```
micgro42, line 12: I'm somewhat torn between adding that code now and removing it again in a few days. But it is already written, so we might as well add it.

micgro42 (approved):
Looks good. Should we rename `itemId` to `id` in the interface?

tests/unit/data-access/MwApiItemSearcher.test.ts
lucaswerkmeister: I don’t follow, why would we remove it again?

src/data-access/DevItemSearcher.ts
```diff
@@ -0,0 +1,48 @@
...
+						value: `fake description for ${itemId}`,
+					},
+				},
+			} ];
+		}
+
+		const response = await fetch( 'https://www.wikidata.org/w/api.php?' + new URLSearchParams( {
+			...commonParams,
+			search: searchTerm,
+			language: 'en',
```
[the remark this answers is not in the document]
lucaswerkmeister: It isn’t fully ignored – compare:

https://www.wikidata.org/w/api.php?action=wbsearchentities&search=Berlim&language=de
https://www.wikidata.org/w/api.php?action=wbsearchentities&search=Berlim&language=pt

src/data-access/ItemSearcher.ts
```diff
@@ -0,0 +1,19 @@
+export interface SearchedItemOption {
+	itemId: string;
```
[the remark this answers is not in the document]
lucaswerkmeister: I’m fine with renaming it, but I don’t think it would reduce the amount of “work”, since I’d still put something like `id: result.id` in `processResponse`. (I wouldn’t just return the response directly, even though, with the `display` changes merged, it should be compatible.)

tests/unit/data-access/MwApiItemSearcher.test.ts
micgro42: removing in the sense of basically completely rewriting, because we're targeting the new version of the API response that has the `display` key for us to use.

src/data-access/ItemSearcher.ts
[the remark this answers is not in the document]
micgro42: ok, then let's keep it as it is

src/data-access/DevItemSearcher.ts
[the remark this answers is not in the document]
lucaswerkmeister: Yeah, the `strictlanguage` part might be a bug. Without that parameter, it looks like the search is meant to eventually fall back to terms in any language, so the `language` would mainly be used for ordering, I guess.

PR #91, opened 2022-03-18
Add default parameters to mw.Api()

lucaswerkmeister:
We want to make sure that the API uses the current user interface language, even if that was only set via `?uselang=` and isn’t in the user’s session, and also want to use the modern result format. Also, let’s set the error format already, though for now this is mostly a guess since we haven’t implemented error handling yet.

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -12,8 +12,17 @@ interface InitConfig extends Config {
 }
 
 export default function init( config: InitConfig, mw: MediaWiki ): ComponentPublicInstance {
+	const languageCode = mw.config.get( 'wgUserLanguage' ) as string;
+	const api = new mw.Api( {
+		parameters: {
+			formatversion: 2,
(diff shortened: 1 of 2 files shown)


src/init.ts
```diff
@@ -12,8 +12,17 @@ interface InitConfig extends Config {
 }
 
 export default function init( config: InitConfig, mw: MediaWiki ): ComponentPublicInstance {
+	const languageCode = mw.config.get( 'wgUserLanguage' ) as string;
+	const api = new mw.Api( {
+		parameters: {
+			formatversion: 2,
+			uselang: languageCode,
+			errorformat: 'html',
```
micgro42, line 20: I guess this should be ok, especially when looking at the alternative values here. We can look into this in detail when we actually come to that story.

micgro42 (approved):

PR #94, opened 2022-03-21
Go to newly created Lexeme

micgro42:
Todo:
* [x] comments
* [x] jest test for form submit
* [x] assert routing in Cypress

Bug: T304397

### DIFF
--- a/src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
+++ b/src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
@@ -1,9 +1,5 @@
 import WikiRouter from './WikiRouter';
 import { MwUtilGetUrl } from '@/@types/mediawiki';
-import {
-	inject,
-	InjectionKey,
-} from 'vue';
 
 export default class MediaWikiRouter implements WikiRouter {
 
@@ -33,12 +29,4 @@ export default class MediaWikiRouter implements WikiRouter {
 		return this.getUrl( title );
 	}
 
-}
-
-export const WikiRouterKey: InjectionKey<WikiRouter> = Symbol( 'WikiRouter' );
-
-export function useWikiRouter(): WikiRouter {
-	return inject( WikiRouterKey, () => {
-		throw new Error( 'No WikiRouter provided!' );
-	}, true );
 }
--- a/src/plugins/WikiRouterPlugin/WikiRouter.ts
+++ b/src/plugins/WikiRouterPlugin/WikiRouter.ts
@@ -1,3 +1,13 @@
+import { inject, InjectionKey } from 'vue';
+
 export default interface WikiRouter {
 	goToTitle( title: string ): void;
+}
+
+export const WikiRouterKey: InjectionKey<WikiRouter> = Symbol( 'WikiRouter' );
+
+export function useWikiRouter(): WikiRouter {
+	return inject( WikiRouterKey, () => {
+		throw new Error( 'No WikiRouter provided!' );
+	}, true );
 }
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -12,7 +12,7 @@ import {
 	SET_LEMMA,
 	SET_LEXICAL_CATEGORY,
 } from '@/store/mutations';
-import { useWikiRouter } from '@/plugins/WikiRouterPlugin/MediaWikiRouter';
+import { useWikiRouter } from '@/plugins/WikiRouterPlugin/WikiRouter';
 
 const $messages = useMessages();
 const store = useStore();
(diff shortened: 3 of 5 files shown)


src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
```diff
@@ -0,0 +1,34 @@
...
+
+	private readonly getUrl: MwUtilGetUrl;
+
+	public constructor(
+		getUrl: MwUtilGetUrl,
+	) {
+		this.getUrl = getUrl;
+	}
+
+	public getPageUrl( title: string, params?: Record<string, unknown> ): string {
```
itamargiv, line 18: Does this method need to be public? I can't see where it would be used besides `goToTitle`

src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
```diff
@@ -0,0 +1,34 @@
...
+	inject,
+	InjectionKey,
+} from 'vue';
+
+export default class MediaWikiRouter implements WikiRouter {
+
+	private readonly getUrl: MwUtilGetUrl;
+
+	public constructor(
+		getUrl: MwUtilGetUrl,
```
itamargiv, line 13: While I understand why this is injected here, it is not self-evident why we would need to inject a service that gets / generates a URL for us (and why this is better than just having a base URL as a config instead and constructing the URL ourselves). Therefore, I think it would be beneficial to anyone reading this code in the future if a comment was added to document this dependency.

itamargiv (commented):
All in all, I like this approach of injecting a "WikiRouter" to handle navigation out of the app. I added a couple of comments about things that are not so obvious at first glance, and might require some additional clarification.

src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
micgro42: Not for now, though we will probably need it later. I'll make it private, that way I also can write one less test :raised_hands:

src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
micgro42: That is a good point! Though I feel I'm not at my most eloquent today. I added a first comment, but I'd be happy about ideas for how to improve it

micgro42:
I think this is now also ready for review. Happy to hear any suggestions for improvement :)

micgro42:
I think this is now also ready for review. Happy to hear any suggestions for improvement :)

src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
```diff
@@ -0,0 +1,40 @@
...
+	InjectionKey,
+} from 'vue';
+
+export default class MediaWikiRouter implements WikiRouter {
+
+	private readonly getUrl: MwUtilGetUrl;
+
+	/**
+	 * This uses the standard Mediawiki utility, because that is simpler than constructing it
+	 * ourselves out of multiple pieces of config.
```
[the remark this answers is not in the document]
micgro42: Thanks!

tests/mocks/unusedWikiRouter.ts
```diff
@@ -0,0 +1,7 @@
+import WikiRouter from '@/plugins/WikiRouterPlugin/WikiRouter';
+
+const unusedWikiRouter: WikiRouter = {
+	goToTitle: jest.fn().mockRejectedValue( new Error( 'WikiRouter should not be used in this test' ) ),
```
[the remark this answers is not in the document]
micgro42: yeah, I chose `.mockImplementation`. Though maybe we could indeed just use a plain function :thinking:

src/plugins/WikiRouterPlugin/WikiRouter.ts
```diff
@@ -0,0 +1,3 @@
+export default interface WikiRouter {
+	goToTitle( title: string, params?: Record<string, unknown> ): void;
```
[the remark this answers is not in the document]
micgro42: done

src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+}
+
+export const WikiRouterKey: InjectionKey<WikiRouter> = Symbol( 'WikiRouter' );
+
+export function useWikiRouter(): WikiRouter {
+	return inject( WikiRouterKey, () => {
+		throw new Error( 'No WikiRouter provided!' );
+	}, true );
+}
```
lucaswerkmeister, line 44: Having this in `MediaWikiRouter.ts` feels strange to me… I think it would fit better in `WikiRouter.ts`, WDYT?

lucaswerkmeister (approved):
Tested locally, seems to work fine. One optional comment.

micgro42:
(rebased on top of main)

src/plugins/WikiRouterPlugin/MediaWikiRouter.ts
micgro42: yeah, that makes more sense. I'll move it!

lucaswerkmeister (approved):

PR #97, opened 2022-03-23
Use lookup for lexical category

micgro42:
The main chunk of the work for this was done in #78 for T298139. Now
we're reaping the benefits by being able to reuse all that code.

Bug: T298149


lucaswerkmeister (dismissed):
Works great, just one small thing. (Marking the review as approved anyways because the eslint output is only a warning, not an error, and I’m not sure how we deal with those…)

lucaswerkmeister (approved):

PR #98, opened 2022-03-23
Move config from store to plugin

lucaswerkmeister:
The config isn’t mutable state, so it doesn’t really belong in the store. A plugin that components can use directly seems to make more sense. (Mark the members of the new `Config` interface `readonly` to highlight that they’re not mutable.)

Bug: T303050


micgro42 (approved):
looks great!

PR #99, opened 2022-03-25
Recommend vscode extensions

micgro42:


micgro42:
I still think this is a good idea in principle, but let's revisit it at a later point in time

PR #100, opened 2022-03-25
Fix Lookup keyboard scrolling onwiki

micgro42:
For some reason, scrolling in the lookups with the Keyboard is broken when our app actually runs on mediawiki, but it works fine when it runs as a standalone on Netlify. I'm not yet sure about the cause.


micgro42:
It turned out to be caused by a too old Vue 3 version provided by MediaWiki. See T304717

PR #101, opened 2022-03-25
Error message display

chukarave:
- Create an ErrorMessage component to be used in the New Lexeme Form
- Add errors to the store 
- Add correspondent actions and mutations
- set up DevLexemeCreator to display an error at will

Bug: T304490


micgro42 (approved):

micgro42:
Next steps would then probably be to add the i18n for the "unkown error"-message to the dev messages plugin. But that can happen in an extra PR.

PR #102, opened 2022-03-28
Add wikibaseLexemeTermLanguages to Config interface.

guergana:
Bug: T304496


src/plugins/ConfigPlugin/Config.ts
```diff
@@ -0,0 +1,18 @@
+import {
+	inject,
+	InjectionKey,
+} from 'vue';
+
+export interface Config {
+	readonly licenseUrl: string;
+	readonly licenseName: string;
+	readonly wikibaseLexemeTermLanguages: string;
```
lucaswerkmeister, line 9: This should be a `string[]`, no?

guergana: yes. indeed

tests/unit/main.test.ts
```diff
@@ -16,6 +16,7 @@ describe( 'createAndMount', () => {
 			rootSelector: '#test-app',
 			licenseUrl: '',
 			licenseName: '',
+			wikibaseLexemeTermLanguages: [ 'test' ],
```
lucaswerkmeister, line 19: Maybe use `en` instead of `test`? With `test` it’s not very clear that these should be language codes, IMHO.

lucaswerkmeister (approved):

PR #103, opened 2022-03-29
Make MwApiLexemeCreator return API errors

lucaswerkmeister:
If there are proper HTML errors in the API response, return those so we can show them (at least the first one) in NewLexemeForm. If there aren’t any errors, use the mw.Api()-generated code instead of hard-coding “unknown” (it will probably be “http” in case of a network error, or possibly “ok-but-empty” or “unknown”).

Bug: T304489

### DIFF
--- a/src/data-access/MwApiLexemeCreator.ts
+++ b/src/data-access/MwApiLexemeCreator.ts
@@ -54,7 +54,7 @@ export default class MwApiLexemeCreator implements LexemeCreator {
 			.catch( ( code: string, _?: unknown, result?: ApiResult ): Promise<never> => {
 				let errors: SubmitError[];
 				try {
-					if ( result && Array.isArray( result.errors ) ) {
+					if ( result && result.errors ) {
 						errors = result.errors.map( ( apiError ) => {
 							const submitError: SubmitError = { type: apiError.code };
 							if ( apiError.html ) {
--- a/src/data-access/MwApiLexemeCreator.ts
+++ b/src/data-access/MwApiLexemeCreator.ts
@@ -54,16 +54,21 @@ export default class MwApiLexemeCreator implements LexemeCreator {
 			.catch( ( code: string, _?: unknown, result?: ApiResult ): Promise<never> => {
 				let errors: SubmitError[];
 				try {
-					// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-					errors = result!.errors.map( ( apiError ) => {
-						const submitError: SubmitError = { type: apiError.code };
-						if ( apiError.html ) {
-							submitError.message = apiError.html;
-						}
-						return submitError;
-					} );
-				} catch {
-					errors = [ { type: code } ];
+					if ( result && Array.isArray( result.errors ) ) {
+						errors = result.errors.map( ( apiError ) => {
+							const submitError: SubmitError = { type: apiError.code };
+							if ( apiError.html ) {
+								submitError.message = apiError.html;
+							}
+							return submitError;
+						} );
+					} else {
+						errors = [ { type: code } ];
+					}
+				} catch ( e ) {
+					// eslint-disable-next-line no-console
+					console.error( 'Unexpected API result', result, e );
+					errors = [ { type: 'assertionerror' } ];
 				}
 				return Promise.reject( errors );
 			} );
--- a/src/data-access/MwApiLexemeCreator.ts
+++ b/src/data-access/MwApiLexemeCreator.ts
@@ -1,10 +1,18 @@
 import { MwApi } from '@/@types/mediawiki';
 import LexemeCreator from '@/data-access/LexemeCreator';
+import { SubmitError } from '@/store/RootState';
 
 type WbEditEntityResponse = {
 	entity: {
 		id: string;
 	};
+};
+
+type ApiResult = {
+	errors: {
+		code: string;
+		html: string;
+	}[];
 };
 
 export default class MwApiLexemeCreator implements LexemeCreator {
@@ -37,10 +45,27 @@ export default class MwApiLexemeCreator implements LexemeCreator {
 				language: lexemeLanguageItemId,
 				lexicalCategory: lexicalCategoryItemId,
 			} ),
+			errorformat: 'html',
+			formatversion: 2,
 		} );
 		const response = await this.api.postWithEditToken( params )
-			.catch( () => {
-				return Promise.reject( [ { type: 'unknown' } ] );
+			// jQuery promises can resolve and reject with multiple values;
+			// the third argument is either missing or the result
+			.catch( ( code: string, _?: unknown, result?: ApiResult ): Promise<never> => {
+				let errors: SubmitError[];
+				try {
+					// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+					errors = result!.errors.map( ( apiError ) => {
+						const submitError: SubmitError = { type: apiError.code };
+						if ( apiError.html ) {
+							submitError.message = apiError.html;
+						}
+						return submitError;
+					} );
+				} catch {
+					errors = [ { type: code } ];
+				}
+				return Promise.reject( errors );
 			} );
 
 		return ( response as WbEditEntityResponse ).entity.id;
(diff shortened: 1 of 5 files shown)


tests/unit/data-access/MwApiLexemeCreator.test.ts
```diff
@@ -36,7 +38,64 @@ describe( 'MwApiLexemeCreator', () => {
...
 			assert: 'user',
+			errorformat: 'html',
+			formatversion: 2,
 		} );
 	} );
 
+	describe( 'error handling', () => {
+
+		it( 'maps API errors', () => {
+			const deferred = $.Deferred();
```
itamargiv, line 49: I'm a little confused as to why we need to use jQuery deferred objects here? Can't this be done with promises?

src/data-access/MwApiLexemeCreator.ts
```diff
@@ -37,10 +45,27 @@ export default class MwApiLexemeCreator implements LexemeCreator {
...
+					errors = result!.errors.map( ( apiError ) => {
+						const submitError: SubmitError = { type: apiError.code };
+						if ( apiError.html ) {
+							submitError.message = apiError.html;
+						}
+						return submitError;
+					} );
+				} catch {
+					errors = [ { type: code } ];
+				}
```
itamargiv, line 67: To make sure I understand correctly, we are unsure is result exists (or if results is an array) before we map it, so we use this `try catch` to pattern match on result? Shouldn't we add more explicit logic to check these conditions? Or am I missing a different scenario where the error mapping would throw?

itamargiv (commented):
It's looking really good, I have a few questions about the code. And generally speaking, I think I'd prefer to avoid pulling in jQuery as a dependency if we can.

src/data-access/MwApiLexemeCreator.ts
```diff
@@ -37,10 +45,27 @@ export default class MwApiLexemeCreator implements LexemeCreator {
 				language: lexemeLanguageItemId,
 				lexicalCategory: lexicalCategoryItemId,
 			} ),
+			errorformat: 'html',
+			formatversion: 2,
```
micgro42, line 49: Aren't these duplicating what we have in init.ts, the `new mw.API()` constructor?

src/data-access/MwApiLexemeCreator.ts
```diff
@@ -37,10 +45,27 @@ export default class MwApiLexemeCreator implements LexemeCreator {
...
 		const response = await this.api.postWithEditToken( params )
-			.catch( () => {
-				return Promise.reject( [ { type: 'unknown' } ] );
+			// jQuery promises can resolve and reject with multiple values;
+			// the third argument is either missing or the result
+			.catch( ( code: string, _?: unknown, result?: ApiResult ): Promise<never> => {
+				let errors: SubmitError[];
+				try {
+					// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+					errors = result!.errors.map( ( apiError ) => {
```
micgro42, line 58: Can't we just check if `results`  is an object with an `errors` key that holds an array with an `if` condition?
That way we would not need neither the try-catch nor the non-null-assertion here.

src/data-access/MwApiLexemeCreator.ts
lucaswerkmeister: > To make sure I understand correctly, we are unsure if `result` is set (or if it is an array) before we map it, so we use this `try catch` to pattern match on it?

Yes, it might be `undefined` or empty.

> Shouldn't we add more explicit logic to check these conditions?

To me this seems nicer than checking several condition in sequence (is object, has errors, errors are array…).

src/data-access/MwApiLexemeCreator.ts
lucaswerkmeister: Yes, but I thought they should be included here again, since we now rely on both of these parameters – if the code relies on a certain API format, it should also ensure that it’s actually in use, not just hope that the given `api` instance does the right thing.

I could be convinced otherwise, but then the requirements should at least be documented in the class comment, I think.

micgro42: > Yes, but I thought they should be included here again, since we now rely on both of these parameters

Yes, that is a good point. But shouldn't we, by that logic, also include the `uselang` here? We are also depending on it being returned in a specific language.

lucaswerkmeister: I wouldn’t say so – the code here works fine with any `uselang`, whether it’s appropriate to show to users is a different question (I think we don’t even have the language code in question in this class?) – whereas `errorformat` and `formatversion` change the JSON structure: without `errorformat`, there would be a single `error` instead of an `errors` array, and without `formatversion` the `html` key would be `*` instead.

src/data-access/MwApiLexemeCreator.ts
lucaswerkmeister: I guess so, though I’d say there’s still a lot going on in that snippet…

itamargiv: Yeah, we don't need to check if it's an object, only arrays can be mapped in JS. Anything other than an array will return false for that check. As for the logic inside the callback, that can stay the same, it's just replacing the `try... catch...` (which can inadvertently cause this mapping to fail for other reasons that we might not want to so easily catch) with what the actual intent of this snippet is -  map errors from an array if that array is there and provide us with a code otherwise.

lucaswerkmeister: (With “is object” I was referring to `result`, not `result.errors`.)

lucaswerkmeister: We discussed this in a call, I’ve pushed a new commit that shifts to using `if` inside a try/catch (we keep the try/catch to ensure we don’t reject the promise with something other than an array of errors).

tests/unit/data-access/MwApiLexemeCreator.test.ts
lucaswerkmeister: Pushed a commit that uses a mock object instead, we can decide if we like that better or not. (I’d say it turned out less terrible than I feared.)

tests/unit/data-access/MwApiLexemeCreator.test.ts
```diff
@@ -1,7 +1,17 @@
...
 import unusedApi from '../../mocks/unusedApi';
 
+/* Return an object that works a bit like a jQuery.Deferred() rejected with the given args. */
+function mockRejectedDeferred( ...args: unknown[] ): unknown {
+	return {
+		catch<T>( callback: ( ..._: unknown[] ) => PromiseLike<T> ): Promise<T> {
+			return Promise.resolve( callback( ...args ) );
+		},
+	};
+}
```
itamargiv, line 13: Short and sweet! I love it. Thank you :pray:

src/data-access/MwApiLexemeCreator.ts
```diff
@@ -37,10 +45,32 @@ export default class MwApiLexemeCreator implements LexemeCreator {
...
 		} );
 		const response = await this.api.postWithEditToken( params )
-			.catch( () => {
-				return Promise.reject( [ { type: 'unknown' } ] );
+			// jQuery promises can resolve and reject with multiple values;
+			// the third argument is either missing or the result
+			.catch( ( code: string, _?: unknown, result?: ApiResult ): Promise<never> => {
+				let errors: SubmitError[];
+				try {
+					if ( result && Array.isArray( result.errors ) ) {
```
lucaswerkmeister, line 57: Maybe this should just be `result && result.errors`? If `result.errors` exists, but is not an array, then I feel like this should go through the `catch` block (`type: 'assertionerror'`) rather than the `else` block (`type: code`) – after all, the API has responded in a highly unexpected way in that case.

itamargiv: Yeah, that makes sense to me.

lucaswerkmeister: Alright, thanks. Pushed.

itamargiv (approved):
Looks good yo me, let's get this merged.

Dog LGTM

PR #105, opened 2022-03-29
Create the spelling variant lookup field

guergana:
Bug: T304500

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,7 +17,7 @@
 				rootSelector: '#app',
 				licenseUrl: '',
 				licenseName: '',
-				wikibaseLexemeTermLanguages: [ 'en', 'en-ca', 'es', 'aa', 'bcc', 'ca', 'de', 'ff', 'ga', 'ha-arab' ],
+				wikibaseLexemeTermLanguages: [ 'en', 'en-ca', 'en-ca', 'en-gb', 'enm', 'es', 'est', 'esp', 'esk', 'aa', 'bcc', 'ca', 'de', 'ff', 'ga', 'ha-arab' ],
 			};
 			const services = {
 				itemSearcher: new DevItemSearcher(),
--- a/index.html
+++ b/index.html
@@ -17,7 +17,7 @@
 				rootSelector: '#app',
 				licenseUrl: '',
 				licenseName: '',
-				wikibaseLexemeTermLanguages: [ 'en', 'en-ca', 'en-ca', 'en-gb', 'enm', 'es', 'est', 'esp', 'esk', 'aa', 'bcc', 'ca', 'de', 'ff', 'ga', 'ha-arab' ],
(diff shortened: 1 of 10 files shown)


lucaswerkmeister (commented):
Trying it out locally, it looks alright. I notice some things that bother me about the usability, but I’m not sure how many of them need to be addressed in this PR. In any case, I suggest we split out the store part (add spelling variant to state, add mutation) into a separate PR, so that this part can be merged quickly.

The usability things I’ve noticed:
- By default, the input is just empty. On Special:NewLexeme, it’s a dropdown, with a button you can click to see all the options – that seems more convenient to me.
- The searching is case sensitive – if I didn’t notice the placeholder, and typed “English”, then I will never see results, since after the “E” there are already no matching language codes left.
- Even when the input is a full language code that exists, it’s not always the first option – e.g. typing ”ng” will produce `ang` before `ng` (followed by `shi-tfng`, `ccp-beng` etc.). I think in Special:NewLexeme the matching condition is “starts with the input” instead of “contains the input” (i.e. `ang` wouldn’t even match), which might be better.
- Pressing enter to try to select a language code instead tries to submit the whole form.
- The lookup doesn’t go away when the input loses focus, even when you click somewhere else or focus another input field.

guergana:
I noticed what you mentioned about the lookup not having an arrow to see the available options. The parent ticket suggests using the [Lookup component](https://wmde.github.io/wikit/?path=/docs/vue_lookup-lookup--all), which doesn't have an arrow. I was not there when this ticket was written, so I thought this had been already discussed. For this to be possible the Lookup component would need to be modified. Another option is to use the [Dropdown component](https://wmde.github.io/wikit/?path=/docs/vue_dropdown--basic) but it is not searchable. This probably needs to be discussed with the team. 

In the meantime I will address the rest of the issues you mention.

index.html
```diff
@@ -17,6 +17,7 @@
 				rootSelector: '#app',
 				licenseUrl: '',
 				licenseName: '',
+				wikibaseLexemeTermLanguages: [ 'en' ],
```
micgro42, line 20: Could we have a lot more here? Maybe like 10 or so.

Then https://ed85c311d1a49ef3e851d47e436a740779b1--new-lexeme-special-page.netlify.app/ is more useful

micgro42 (changes requested):
It looks good overall, but I think I would like to see an assertion about selecting a new spelling variant in tests/integration/NewLexemeForm.test.ts

index.html
guergana: you are right. will add a few

guergana:
> It looks good overall, but I think I would like to see an assertion about selecting a new spelling variant in tests/integration/NewLexemeForm.test.ts

It's been added. :)

micgro42 (commented):
I think this can be merged, though there are a couple of follow-up tasks still open before this story is done:
* [ ] actually make use of the selected language code when creating the new Lexeme
* [ ] showing and selecting based on the language name

And two bugs with an overarching scope:

* [ ] figuring out and adjusting the behavior on what happens when pressing "Enter" => T305359
* [ ] figuring out and fixing the menu behavior on focus gain/loss on-wiki

PR #106, opened 2022-03-30
Avoid console.error() in test and add extra test

lucaswerkmeister:
The first test for the `CREATE_LEXEME` action was logging an error to the console (though that doesn’t make the test fail), complaining about the `CLEAR_ERRORS` mutation being missing. Fix that, and also add another test checking that the action clears and adds errors as expected.

The new test includes this line of code:

```js
await expect( store.dispatch( CREATE_LEXEME ) ).rejects.toBe( null );
```

In my opinion, we shouldn’t actually care about *what* the promise rejects with – however, `await expects( … ).rejects;` on its own doesn’t work: you have to perform some match on the rejected value, otherwise it counts as an unhandled promise rejection. I couldn’t find a “just accept anything” matcher that works here, so let’s just assert null, I guess.

### DIFF
--- a/tests/unit/store/actions.test.ts
+++ b/tests/unit/store/actions.test.ts
@@ -1,6 +1,10 @@
 import createActions, { CREATE_LEXEME } from '@/store/actions';
 import LexemeCreator from '@/data-access/LexemeCreator';
-import RootState from '@/store/RootState';
+import {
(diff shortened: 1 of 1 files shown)


chukarave (approved):
works well for me locally

PR #124, opened 2022-04-04
Add linting for BEM-style classes in scss

micgro42:
This effectively extends the functionality of the selector-class-pattern
rule, and so that rule can be dropped from the configuration.

From the other rule-sets mentioned in T301694,
stylelint-selector-bem-pattern has much more complex configuration for
little apparent benefit. And Vue-BEM adds a new custom directive that
also seems like too much complexity for what little of a problem we here
have.

Thus, this uses https://github.com/merkle-open/stylelint-bem

Bug: T301694

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -4,17 +4,15 @@
 		"stylelint-config-standard-scss",
 		"stylelint-config-recommended-vue/scss"
 	],
-	"plugins": ["stylelint-use-logical-spec"],
+	"plugins": ["stylelint-use-logical-spec", "@namics/stylelint-bem"],
 	"rules": {
+		"plugin/stylelint-bem-namics": {
+			"patternPrefixes": [],
+			"helperPrefixes": [],
+			"namespaces": ["wbl-snl-"]
+		},
 		"indentation": "tab",
 		"selector-max-id": 1,
-		"liberty/use-logical-spec": true,
-		"selector-class-pattern": [
-			"^wbl-snl-[-_a-z0-9]+$",
-			{
-				"resolveNestedSelectors": true,
-				"message": "Selector class names should be lowercase and prefixed with wbl-snl-*."
-			}
-		]
+		"liberty/use-logical-spec": true
 	}
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
 				"vuex": "^4.0.2"
 			},
 			"devDependencies": {
+				"@namics/stylelint-bem": "^7.0.0",
 				"@rushstack/eslint-patch": "^1.1.0",
 				"@types/jest": "^27.4.0",
 				"@types/lodash": "^4.14.180",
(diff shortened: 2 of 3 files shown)


micgro42:
> > > I’m not sure what this actually changes… it doesn’t seem to lint `class=""` attributes in the `<template>` section, which I thought was what we wanted in [T301694](https://phabricator.wikimedia.org/T301694)?
> > 
> > 
> > No, it is an update to stylelint, so it lints the classes we use in the `<style>` block.
> 
> Okay, but we’re already linting the classes in the `<style>` block…

Yes, but so far this only covered a rather general regex of "must be prefixed with `wbl-snl-` and after that every combination of `a-z0-9_-` is fair game. Now it enforces the BEM rules somewhat more strictly, e.g., `.wbl-snl-foo_bar` would be an error, as would be `.wbl-snl-foo--bar__baz`.

Not that it really is a big deal, so far we barely use any styling at all. Though this might change in the next milestone.

micgro42:
rebased.

micgro42:
> > Yes, but so far this only covered a rather general regex of "must be prefixed with `wbl-snl-` and after that every combination of `a-z0-9_-` is fair game". Now it enforces the BEM rules somewhat more strictly, e.g., `.wbl-snl-foo_bar` would be an error, as would be `.wbl-snl-foo--bar__baz`.
> 
> That sounds good, but unrelated to what I thought [T301694](https://phabricator.wikimedia.org/T301694) was about – my understanding of the task was that we want linting for the `class=""` attributes, since (because we don’t have much styling yet) most of these classes don’t have a corresponding selector in a `<style>` block, which means that they don’t get linted at all.

For me, the core motivation still comes through: the classes that we _use_ follow BEM. I think, we only use classes for styling in production code (and not selecting and modifying elements, like with jQuery).

I agree that I would also have preferred to have a _linting_ of classes already in the template. But a cursory search didn't surface any good candidates.

There are helper packages like https://www.npmjs.com/package/@vuebits/bem and https://github.com/mlturner88/vue-simple-bem that offer utilities to _create_ BEM classes, but I feel that is currently overkill for our use-case.

Since we do get most of the desired benefit out of the stylelint plugin and there is no good candidate for linting the classes in templates, I think the remaining options would be to write such a linter ourselves or just use only the stylelint plugin. I feel that for this little additional benefit and it being a 1 story point task, writing it ourselves seems out of scope.

That being said, I would be ok with not closing the Phabricator task after this PR is merged, if we want to keep our eyes open for a better solution in the future.

Also, maybe writing such a linter could be ok, if we also want to use it in other projects and make it available as an Open Source project to the WMF/Community/The World. Maybe that could be a Ginger Beer'n'Cake or Hackathon project?

itamargiv:
My 2 cent: I think this plugin does exactly what we want it to do. The BEM convention was designed to organize stylesheets, not HTML elements. And the plugin does exactly that, checks stylesheets to ensure they are organized with the BEM convention. If we are not intending to use BEM classes for styling an element, then there should be no requirement for the class name to be BEM compliant. Since we can add multiple classes to an element, do all of them have to be BEM?

lucaswerkmeister:
I don’t care that much about BEM, I think, but IMHO all the classes in our HTML (even the ones not used in CSS) should start with a constant prefix – to avoid collisions with other styles on-wiki. That’s what I thought we wanted to lint for – but we can also decline the task and keep checking it manually in code review instead.

micgro42:
> I don’t care that much about BEM, I think, but IMHO all the classes in our HTML (even the ones not used in CSS) should start with a constant prefix – to avoid collisions with other styles on-wiki. That’s what I thought we wanted to lint for

But, none of that is actually in the words of the task? :upside_down_face: 

Also, at least for the classes in CSS, we do that already and the change proposed here keeps doing that. For classes in the template, we keep having to check for it manually, yes.

lucaswerkmeister:
> But, none of that is actually in the words of the task? upside_down_face

“A linter or linters are set to check that classes we add to **HTML and** CSS follow the BEM convention” (emphasis added). Otherwise I’m mainly basing this on the discussions referenced in the first paragraph (“During sprint 1 we (the devs) noted that we might want to follow some convention regarding CSS classes … we also concluded that it would be too much overhead for developers to check adherence to BEM rules in each PR”), which was somewhere on GitHub but I don’t remember where.

micgro42:
I think it was here: https://github.com/wmde/new-lexeme-special-page/pull/33#discussion_r805947489

lucaswerkmeister:
I’ve slightly updated T301694 to remove the HTML part, and created T305707 for what I had in mind instead. (I think the other discussion I was thinking of might have been https://github.com/wmde/new-lexeme-special-page/pull/54#pullrequestreview-888941391.)

lucaswerkmeister (approved):
I personally don’t see much added value in this, but we can merge it (after rebasing, probably).

lucaswerkmeister (approved):

micgro42:
> I’ve slightly updated [T301694](https://phabricator.wikimedia.org/T301694) to remove the HTML part, and created [T305707](https://phabricator.wikimedia.org/T305707) for what I had in mind instead. (I think the other discussion I was thinking of might have been [#54 (review)](https://github.com/wmde/new-lexeme-special-page/pull/54#pullrequestreview-888941391).)

Cool, thanks! I think the thing in T305707 should be quite doable in a day-project with code based upon a plugin we use in wikit for asserting that every component has a `wikit` class on its root.

PR #125, opened 2022-04-04
Upgrade dependencies with npm upgrade

micgro42:
This includes updates to prettier and vue-tsc beyond the semver range allowed in package.json. Both don't affect us.

This should close this month's dependabot PR's regarding the package.json dependencies.

### DIFF
--- a/cypress/fixtures/example.json
+++ b/cypress/fixtures/example.json
@@ -1,5 +1,0 @@
-{
-	"name": "Using fixtures to represent data",
-	"email": "hello@cypress.io",
-	"body": "Fixtures are a great way to mock data for responses to routes"
-}


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+	"name": "Using fixtures to represent data",
+	"email": "hello@cypress.io",
+	"body": "Fixtures are a great way to mock data for responses to routes"
+}
```
lucaswerkmeister, line 5: We probably don’t want to add this file?

micgro42: ugh, I should probably add this to `.gitignore` Cypress is creating it automatically

micgro42: Done

lucaswerkmeister (approved):

PR #127, opened 2022-04-05
Use new Spelling Variant Lookup when creating new Lexeme

micgro42:
Bug: T304500

### DIFF
--- a/tests/integration/NewLexemeForm.test.ts
+++ b/tests/integration/NewLexemeForm.test.ts
@@ -76,7 +76,7 @@ describe( 'NewLexemeForm', () => {
 			global: {
 				plugins: [ testStore ],
 				provide: {
-					[ ConfigKey as symbol ]: { wikibaseLexemeTermLanguages: [ 'en' ] },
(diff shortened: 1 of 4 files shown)


tests/integration/NewLexemeForm.test.ts
```diff
@@ -90,13 +90,17 @@ describe( 'NewLexemeForm', () => {
 		await languageInput.setValue( '=Q123' );
 		await wrapper.find( '.wbl-snl-language-lookup .wikit-OptionsMenu__item' ).trigger( 'click' );
 
+		const spellingVariantInput = wrapper.find( '.wbl-snl-spelling-variant-lookup input' );
+		await spellingVariantInput.setValue( 'en' );
```
lucaswerkmeister, line 94: Maybe a comment here would be helpful? Something like `// will actually select en-gb, en is not in wikibaseLexemeTermLanguages`

lucaswerkmeister (dismissed):
LGTM, one suggestion

tests/integration/NewLexemeForm.test.ts
micgro42: Done.

lucaswerkmeister (approved):

PR #129, opened 2022-04-07
Add repository to get the LanguageCode from Wikidata

micgro42:
Todo:
- [ ] error handling?

Bug: T305539


micgro42:
Not yet sure how to best handle errors here, but otherwise ready for review

src/data-access/MwApiLangCodeRetriever.ts
```diff
@@ -0,0 +1,21 @@
+import { MwApi } from '@/@types/mediawiki';
+import LangCodeRetriever from './LangCodeRetriever';
+import { processWbGetClaimsResponse, WbGetClaimsResponse } from './apiBaseLangCodeRetriever';
+
+export default class MwApiLangCodeRetriever implements LangCodeRetriever {
+
+	public constructor( private api: MwApi, private languageCodeProperty: string ) {
```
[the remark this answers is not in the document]
micgro42: Yep, Typescript magic. Not convinced by it myself yet, but I thought I try it out. Apparently they are called "parameter properties": https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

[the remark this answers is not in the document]
micgro42: Yeah, but also I think I should actually assert that the languageCodeProperty is actually a string and not `null` because of a config error. So, I'll probably move to the default style again, at least for this case.

src/data-access/DevLangCodeRetriever.ts
```diff
@@ -0,0 +1,17 @@
+import { processWbGetClaimsResponse, WbGetClaimsResponse } from './apiBaseLangCodeRetriever';
```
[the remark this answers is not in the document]
micgro42: true. will fix

src/data-access/apiBaseLangCodeRetriever.ts
```diff
@@ -0,0 +1,34 @@
+import { DataValue, Statement, StatementMap } from '@wmde/wikibase-datamodel-types';
+
+export type WbGetClaimsResponse = {
+	claims: StatementMap;
+};
+
+function getValueFromClaim( claim: Statement ): string | false {
```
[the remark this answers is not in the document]
micgro42: can do

src/data-access/MwApiLangCodeRetriever.ts
```diff
@@ -0,0 +1,21 @@
...
+	}
+
+	public async getLanguageCodeFromItem( itemId: string ): Promise<string | null | false> {
+		// TODO handle errors
```
micgro42, line 11: I feel we can do _something_ here to handle errors. Probably would do a try-catch around the `await this.api.get(...)` and in the catch clause log the error to `console.error()` (maybe with a prefix or something?) and then return false, as to let the App recover from, for example, network problems.

Thoughts?

src/data-access/apiBaseLangCodeRetriever.ts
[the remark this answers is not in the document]
micgro42: Done.

src/data-access/MwApiLangCodeRetriever.ts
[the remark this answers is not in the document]
micgro42: Done.

src/data-access/DevLangCodeRetriever.ts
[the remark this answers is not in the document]
micgro42: Done.

src/data-access/apiBasedLangCodeRetriever.ts
```diff
@@ -4,11 +4,15 @@ export type WbGetClaimsResponse = {
 	claims: StatementMap;
 };
 
-function getValueFromClaim( claim: Statement ): string | false {
+function getStringFromClaim( claim: Statement ): string | false {
```
lucaswerkmeister, line 7: Sorry, I didn’t notice this before – I would also avoid the term “claim” and just use “statement” throughout. (Same goes for the `( claim ) =>` filter functions below.)

src/data-access/LangCodeRetriever.ts
```diff
@@ -0,0 +1,10 @@
+export default interface LangCodeRetriever {
+
+	/*
+	 * Returns the language code associated with the Item,
+	 * which may or may not be a valid language code, validation needs to be done by the caller
+	 * returns `null` if the Item does not have a language code associated with it
+	 * returns `false` if the Item has novalue or somevalue for the language code property
```
lucaswerkmeister, line 7: I think there should be some punctuation here, otherwise it’s going to look weird in some IDEs (“…associated with it returns `false` if…”).
```suggestion
	 * Returns the language code associated with the Item,
	 * which may or may not be a valid language code, validation needs to be done by the caller.
	 * Returns `null` if the Item does not have a language code associated with it.
	 * Returns `false` if the Item has novalue or somevalue for the language code property.
```

src/data-access/MwApiLangCodeRetriever.ts
lucaswerkmeister: I would implement that as a `.catch()` call instead of `try/catch` syntax, so that we can access all of the values that jQuery promises can reject with (for API errors, the first value is the least useful one). Something like:

```js
.catch( function () {
    console.error( 'API error while getting language code', arguments );
    return false;
} );
```

(Maybe also `console.warn()` instead of `console.error()`, not sure…) But in general that sounds okay to me.

micgro42: sounds good, will do!

src/data-access/apiBasedLangCodeRetriever.ts
micgro42: mh, right. I think I remember somthing about deprecating the term "claim" and replacing it with "statement". I'll change it where I can

src/data-access/MwApiLangCodeRetriever.ts
micgro42: Ok, I have a version up that I think should work

src/data-access/apiBasedLangCodeRetriever.ts
micgro42: I think, I've replaced all the ones that I can

tests/unit/data-access/MwApiLangCodeRetriever.test.ts
```diff
@@ -0,0 +1,118 @@
...
+		api.get = jest.fn().mockResolvedValue( {
+			claims: {},
+		} );
+
+		const sut = new MwApiLangCodeRetriever( api, 'P218' );
+
+		expect( await sut.getLanguageCodeFromItem( 'Q5' ) ).toBe( null );
+	} );
+
+	it( 'picks the preffered rank statement if available', async () => {
```
lucaswerkmeister, line 18:
```suggestion
	it( 'picks the preferred rank statement if available', async () => {
```

tests/unit/data-access/MwApiLangCodeRetriever.test.ts
```diff
@@ -0,0 +1,118 @@
...
+				],
+			},
+		} );
+
+		const sut = new MwApiLangCodeRetriever( api, 'P218' );
+
+		expect( await sut.getLanguageCodeFromItem( 'Q9301' ) ).toBe( false );
+	} );
+
+	it( 'returns the language code form the normal ranked statement if there is no preferred one', async () => {
```
lucaswerkmeister, line 59:
```suggestion
	it( 'returns the language code from the normal ranked statement if there is no preferred one', async () => {
```

lucaswerkmeister (commented):
LGTM apart from two typos, but I haven’t tested it manually yet, and am not sure how to do that either. (Maybe it’s not necessary at this stage.)

lucaswerkmeister (approved):

micgro42:
🙌

PR #130, opened 2022-04-11
Create Action to infer LanguageCode

micgro42:
Todo:
* [x] tests
* [x] validation
* [x] reacting to it

Cypress test will be updated in a separate commit. (In this commit, it probably asks for the language code of [September (Q123)](https://www.wikidata.org/wiki/Q123), which isn’t very nice but doesn’t need to block merging this.) Showing a warning will also be done separately.

Bug: T305542


micgro42:
(rebased on top of `main` after merging #131)

PR #131, opened 2022-04-11
Use language codes via plugin

micgro42:
This will be useful when validating the language code that we get from
the API for the language item. See #130

Bug: T305542


lucaswerkmeister (approved):

lucaswerkmeister:
Hrm, integration tests need an update.

lucaswerkmeister (approved):

micgro42:
thanks for fixing it!

/me merges

PR #139, opened 2022-04-14
Add&Adjust Cypress tests for language code inference

micgro42:
Bug: T305542

### DIFF
--- a/cypress/integration/NewLexemeForm.spec.js
+++ b/cypress/integration/NewLexemeForm.spec.js
@@ -19,7 +19,7 @@ describe( 'NewLexemeForm', () => {
 						datatype: 'external-id',
 					},
 					type: 'statement',
-					id: 'Q9301$6e2ffdaf-4ec3-055f-7165-4849df92ffde',
+					id: 'Q123$c5c5b451-38a0-4d7c-99b9-af4fdf893d08',
 					rank: 'normal',
 				} ],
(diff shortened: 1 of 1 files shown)


cypress/integration/NewLexemeForm.spec.js
```diff
@@ -1,21 +1,78 @@
...
+						property: 'P218',
+						hash: 'faa942355921695b9fa30805f0bb66bc7ed5a433',
+						datavalue: {
+							value: 'en-gb',
+							type: 'string',
+						},
+						datatype: 'external-id',
+					},
+					type: 'statement',
+					id: 'Q9301$6e2ffdaf-4ec3-055f-7165-4849df92ffde',
```
lucaswerkmeister, line 22: This statement ID doesn’t match the language item ID below, which I think could cause confusion later… I suggest we invent a fresh statement ID.
```suggestion
					id: 'Q123$c5c5b451-38a0-4d7c-99b9-af4fdf893d08',
```

lucaswerkmeister (commented):
looks good, just one suggestion

cypress/integration/NewLexemeForm.spec.js
micgro42: Fixed thanks.

lucaswerkmeister (approved):

PR #140, opened 2022-04-14
Don't crash if no languageCodeProperty configured

micgro42:
Especially in CI, there can be no language code property configured. The
app should still work in that case.

Bug: T306170

### DIFF
--- a/src/data-access/MwApiLangCodeRetriever.ts
+++ b/src/data-access/MwApiLangCodeRetriever.ts
@@ -5,17 +5,18 @@ import { processWbGetClaimsResponse, WbGetClaimsResponse } from './apiBasedLangC
 export default class MwApiLangCodeRetriever implements LangCodeRetriever {
 
 	private readonly api: MwApi;
-	private readonly languageCodeProperty: string;
+	private readonly languageCodeProperty: string | null;
 
-	public constructor( api: MwApi, languageCodeProperty: string ) {
-		if ( typeof languageCodeProperty !== 'string' ) {
-			throw new Error( `Expected languageCodeProperty to be propertyId but received ${languageCodeProperty} instead!` );
-		}
+	public constructor( api: MwApi, languageCodeProperty: string | null ) {
 		this.api = api;
 		this.languageCodeProperty = languageCodeProperty;
 	}
 
 	public async getLanguageCodeFromItem( itemId: string ): Promise<string | null | false> {
+		if ( !this.languageCodeProperty ) {
+			return null;
+		}
+
 		const response = await this.api.get( {
 			action: 'wbgetclaims',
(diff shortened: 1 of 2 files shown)


src/data-access/MwApiLangCodeRetriever.ts
```diff
@@ -5,17 +5,22 @@ import { processWbGetClaimsResponse, WbGetClaimsResponse } from './apiBasedLangC
...
-	private readonly languageCodeProperty: string;
+	private readonly languageCodeProperty: string | null;
 
-	public constructor( api: MwApi, languageCodeProperty: string ) {
-		if ( typeof languageCodeProperty !== 'string' ) {
-			throw new Error( `Expected languageCodeProperty to be propertyId but received ${languageCodeProperty} instead!` );
+	public constructor( api: MwApi, languageCodeProperty: string | null ) {
+		if ( !languageCodeProperty ) {
+			// eslint-disable-next-line no-console
+			console.warn( 'LexemeLanguageCodePropertyId seems not configured, language code inference will not work.' );
```
lucaswerkmeister, line 13: I don’t think we should warn in this case, it’s not like end users of the wiki can do much about it unless they also happen to be sysadmins.

src/data-access/MwApiLangCodeRetriever.ts
```diff
@@ -5,17 +5,22 @@ import { processWbGetClaimsResponse, WbGetClaimsResponse } from './apiBasedLangC
...
 		}
 		this.api = api;
 		this.languageCodeProperty = languageCodeProperty;
 	}
 
 	public async getLanguageCodeFromItem( itemId: string ): Promise<string | null | false> {
+		if ( !this.languageCodeProperty ) {
+			return null;
```
lucaswerkmeister, line 21: Should we update the documentation in `LangCodeRetriever` for this case? I think “Returns `null` if _there is_ no associated language code” (instead of “…if _the item has_ no…”) could work. (In that case we could change the end to “including missing or novalue statements”, or we could just leave that out entirely.)

src/data-access/MwApiLangCodeRetriever.ts
micgro42: mh, to me this feels like a deprecation warning, but since we can't enable it conditionally, we should probably remove it indeed.

src/data-access/MwApiLangCodeRetriever.ts
micgro42: Done.

lucaswerkmeister: ah, right, IIRC that line was at 99 characters or so :D

lucaswerkmeister (approved):

PR #141, opened 2022-04-21
Always handle change in the Language input

micgro42:
This if-wrapper was probably just left over from development, but we
also want to remove the selected language value from the store when it
is deselected in the Lookup.

### DIFF
--- a/tests/integration/NewLexemeForm.test.ts
+++ b/tests/integration/NewLexemeForm.test.ts
@@ -77,11 +77,11 @@ describe( 'NewLexemeForm', () => {
 		await wrapper.find( '.wbl-snl-language-lookup .wikit-OptionsMenu__item' ).trigger( 'click' );
 
 		expect( wrapper.find( '.wbl-snl-spelling-variant-lookup' ).exists() ).toBe( false );
-
 		expect( testStore.state.language ).toBe( 'Q123' );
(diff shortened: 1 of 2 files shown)


tests/integration/NewLexemeForm.test.ts
```diff
@@ -80,6 +80,12 @@ describe( 'NewLexemeForm', () => {
 
 		expect( testStore.state.language ).toBe( 'Q123' );
 		expect( testStore.state.languageCodeFromLanguageItem ).toBe( 'de' );
+
+		await languageLookup.setValue( '=Q12' );
+
+		expect( wrapper.find( '.wbl-snl-spelling-variant-lookup' ).exists() ).toBe( false );
+		expect( testStore.state.language ).toBe( null );
+		expect( testStore.state.languageCodeFromLanguageItem ).toBe( undefined );
```
lucaswerkmeister, line 88: I don’t understand what’s happening here… shouldn’t the `language` now be `Q12` and the `languageCodeFromLanguageItem` still be `de`?

lucaswerkmeister (approved):

tests/integration/NewLexemeForm.test.ts
micgro42: Looks good, thanks!

PR #142, opened 2022-04-21
Show and Filter Language Names in Spelling Variant Lookup

micgro42:
Todo:
* [x] more tests for filtering
* [x] better naming in LangCodeProvider?

Bug: T306515

### DIFF
--- a/src/components/SpellingVariantInput.vue
+++ b/src/components/SpellingVariantInput.vue
@@ -18,10 +18,11 @@ interface WikitMenuItem {
 const props = defineProps<Props>();
 
 const languageCodesProvider = useLanguageCodesProvider();
+const messages = useMessages();
 
 const wbLexemeTermLanguages = languageCodesProvider.getLanguageCodes().map(
 	( [ code, name ]: [ string, string ] ) => ( {
-		label: `${name} (${code})`,
+		label: messages.getUnescaped( 'wikibase-lexeme-lemma-language-option', name, code ),
 		value: code,
 		description: '',
 	} ),
@@ -63,7 +64,6 @@ const onOptionSelected = ( value: unknown ) => {
 	emit( 'update:modelValue', selectedValue );
 };
 
-const messages = useMessages();
 </script>
 
 <script lang="ts">
--- a/src/components/SpellingVariantInput.vue
+++ b/src/components/SpellingVariantInput.vue
@@ -20,12 +20,15 @@ const props = defineProps<Props>();
 const languageCodesProvider = useLanguageCodesProvider();
 const messages = useMessages();
 
-const wbLexemeTermLanguages = languageCodesProvider.getLanguageCodes().map(
-	( [ code, name ]: [ string, string ] ) => ( {
-		label: messages.getUnescaped( 'wikibase-lexeme-lemma-language-option', name, code ),
-		value: code,
-		description: '',
-	} ),
+const wbLexemeTermLanguages: WikitMenuItem[] = [];
+languageCodesProvider.getLanguages().forEach(
+	( name, code ) => {
+		wbLexemeTermLanguages.push( {
+			label: messages.getUnescaped( 'wikibase-lexeme-lemma-language-option', name, code ),
+			value: code,
+			description: '',
+		} );
+	},
 );
 
 const menuItems = ref( [] as WikitMenuItem[] );
--- a/src/components/SpellingVariantInput.vue
+++ b/src/components/SpellingVariantInput.vue
@@ -11,16 +11,20 @@ interface Props {
 interface WikitMenuItem {
 	label: string;
 	description: string;
+	value: string;
 }
 
 const props = defineProps<Props>();
 
 const languageCodesProvider = useLanguageCodesProvider();
 
-const wbLexemeTermLanguages = languageCodesProvider.getLanguageCodes().map( ( lang ) => ( {
-	label: lang,
-	description: '',
-} ) );
+const wbLexemeTermLanguages = languageCodesProvider.getLanguageCodes().map(
+	( [ code, name ]: [ string, string ] ) => ( {
+		label: `${name} (${code})`,
+		value: code,
+		description: '',
+	} ),
+);
 
 const menuItems = ref( [] as WikitMenuItem[] );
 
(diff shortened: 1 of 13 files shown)


src/components/SpellingVariantInput.vue
```diff
@@ -39,7 +43,8 @@ const onSearchInput = ( inputValue: string ) => {
 	}
 
 	menuItems.value = wbLexemeTermLanguages.filter(
-		( lang ) => lang.label.startsWith( lowerCaseInputValue ) );
+		( lang ) => ( new RegExp( `\\b${inputValue}`, 'i' ) ).test( lang.label ),
```
lucaswerkmeister, line 46: I think we should escape the `inputValue` here; thankfully, lodash has an [escapeRegExp](https://lodash.com/docs/4.17.15#escapeRegExp) function (found via [StackOverflow](https://stackoverflow.com/q/3446170/1420237)).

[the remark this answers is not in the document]
lucaswerkmeister: To me it makes more sense here.

src/data-access/LanguageCodesProvider.ts
```diff
@@ -1,18 +1,23 @@
...
-	) {}
+		validLanguageCodes: Record<string, string>,
+	) {
+		this.validLanguageCodes = new Map( Object.entries( validLanguageCodes ) );
+	}
 
-	public getLanguageCodes(): readonly string[] {
-		return this.validLanguageCodes;
+	public getLanguageCodes(): [string, string][] {
+		return Array.from( this.validLanguageCodes.entries() );
```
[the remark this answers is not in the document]
micgro42: improved with spread operator. Still strange that there is no .map on iterators.

src/components/SpellingVariantInput.vue
micgro42: It's not like they would gain anything from messing with that regex, but it doesn't hurt either

micgro42: Done.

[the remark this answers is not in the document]
micgro42: Ok thanks, then let's keep it there :+1:

tests/unit/components/SpellingVariantInput.test.ts
```diff
@@ -45,15 +45,72 @@ describe( 'SpellingVariantInput', () => {
...
+					value: 'en-gb',
 				},
 			] );
 		} );
 	} );
+
+	describe( 'suggested options', () => {
+		it.each( [
+			[
+				'Tun would open Tunisia',
```
lucaswerkmeister, line 63:
```suggestion
				'Tun would open Tunisian',
```

tests/unit/components/SpellingVariantInput.test.ts
```diff
@@ -45,15 +45,72 @@ describe( 'SpellingVariantInput', () => {
...
+				[ [ 'en', 'English' ], [ 'aeb-latn', 'Tunisian Arabic (Latin script)' ], [ 'de', 'German' ] ],
+				'Tun',
+				[ 'aeb-latn' ],
+			],
+			[
+				'matching is not fuzzy - tsa does not match Tunisia',
```
lucaswerkmeister, line 69:
```suggestion
				'matching is not fuzzy - tsa does not match Tunisian',
```

tests/unit/components/SpellingVariantInput.test.ts
```diff
@@ -45,15 +45,72 @@ describe( 'SpellingVariantInput', () => {
...
+				[],
+			],
+			[
+				'Matching looks at each word in the list, i.e. Tunisian Arabic would display if arabic entered',
+				[ [ 'en', 'English' ], [ 'aeb-latn', 'Tunisian Arabic (Latin script)' ], [ 'de', 'German' ] ],
+				'arabic',
+				[ 'aeb-latn' ],
+			],
+			[
+				'Matching looks at each word in the list, i.e. Tunisian Arabic would display if arabic entered',
```
lucaswerkmeister, line 81: I assume this is supposed to be a test for looking at the language code?

micgro42: ah yes. I'll fix the message

micgro42: Done

src/components/SpellingVariantInput.vue
```diff
@@ -11,16 +12,20 @@ interface Props {
...
 
 const languageCodesProvider = useLanguageCodesProvider();
 
-const wbLexemeTermLanguages = languageCodesProvider.getLanguageCodes().map( ( lang ) => ( {
-	label: lang,
-	description: '',
-} ) );
+const wbLexemeTermLanguages = languageCodesProvider.getLanguageCodes().map(
+	( [ code, name ]: [ string, string ] ) => ( {
+		label: `${name} (${code})`,
```
lucaswerkmeister, line 24: Hm, should this be localized? ({{optional}}, identical to e.g. authprovider-confirmlink-option)

lucaswerkmeister (commented):
I feel a bit strange about the different formats the language names go through – first they’re an object/record, then a map, then an array of entries. Do we need all of these? ^^

src/components/SpellingVariantInput.vue
micgro42, line 24: This should be the `wikibase-lexeme-lemma-language-option` message #TODO

lucaswerkmeister: Hah, I didn’t realize we already had it in WikibaseLexeme.git ^^

micgro42: Done, though this had somewhat of a ripple effect because we're matching on that label now have to test based on how the messages looks like and how it is replaced. I created #143 for this

micgro42: now used in the newest commit

tests/integration/NewLexemeForm.test.ts
```diff
@@ -30,6 +32,7 @@ describe( 'NewLexemeForm', () => {
 				plugins: [ store ],
 				provide: {
 					[ ConfigKey as symbol ]: {},
+					[ MessagesKey as symbol ]: new Messages( new DevMessagesRepository() ),
 					[ ItemSearchKey as symbol ]: new DevItemSearcher(),
 					[ LanguageCodesProviderKey as symbol ]: {
 						getLanguageCodes: () => [ 'en', 'en-ca', 'de' ],
```
lucaswerkmeister, line 38: Just randomly saw this while looking through the diff – this looks wrong now? (And I guess we should check other `getLanguageCodes` fake implementations too.)

lucaswerkmeister: GitHub won’t let me comment on an unaffected line, but L94/97 looks similar. That’s the only one I found so far, though.

micgro42:
From my side, this _should_ be ready to be merged.

lucaswerkmeister (approved):

PR #143, opened 2022-04-22
Replace parameters in Dev messages

micgro42:
This is mimics what `mw.msg` does. It is needed for testing the
filtering based on the replaced message in the spelling variant options.


lucaswerkmeister (approved):
I think we can merge this, but we should fix the following later:
```
By clicking "Create Lexeme", you agree to the [[&#123;&#123;ns:project&#125;&#125;:Copyrights|terms of use]], and you irrevocably agree to release your contribution under the [ ].
```
The `{{ns:project}}` seems to be double-escaped. (Or we could just change that to `Project` in the dev version of the message, I guess?)

PR #146, opened 2022-04-26
Don’t escape dev messages, only arguments

lucaswerkmeister:
The messages themselves are hard-coded, we know they’re safe; this way, we can also make the links in the shortcopyrightwarning message work.

This means that the NewLexemeForm integration tests now need to supply a full config, otherwise `undefined` ends up being passed into the messages plugin, where it now causes an error.


micgro42 (approved):
The final commit message should mention that this also is closer to what mw.msg does. That is, we're not getting a fully escaped string from that either.

PR #148, opened 2022-04-28
Update netlify command in README.md

lucaswerkmeister:
Since #45 (commit 6dc80b7b9d), we no longer include the `netlify-cli` package in our dependencies, so we need to tell `npx` where the `netlify` command comes from (compare the change to the `deployCommit` command in that commit).

Bug: T302321

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ _**Prerequisits: This repository requires node v16 and up**_
 
 ### The first time you run a deploy
 
-To be able to run the netlify commands locally you will need to run `npx netlify login` once after `npm install` to authenticate and obtain a token.
+To be able to run the netlify commands locally you will need to run `npx -p netlify-cli netlify login` once after `npm install` to authenticate and obtain a token.
 
 ### Every time when doing a deploy
 


lucaswerkmeister:
I should mention that the `deployCommit` command doesn’t actually work for me:

```
$ npm run deployCommit                                                                                                                                                                                                                                      
(snip build)
Deploy path: /var/www/html/wiki1/extensions/WikibaseLexeme/resources/special/new-lexeme/dist                                                                                                                                                                                                                                                               
Deploying to draft URL...                                                                                                                                                                                                                                                                                                                                  
 ›   Warning: JSONHTTPError:  401                                                                                                                                                                                                                                                                                                                          
 ›   Warning:                                                                                                                                                                                                                                                                                                                                              
{                                                                                                                                                                                                                                                                                                                                                          
  "name": "JSONHTTPError",                                                                                                                                                                                                                                                                                                                                 
  "status": 401,                                                                                                                                                                                                                                                                                                                                           
  "json": {                                                                                                                                                                                                                                                                                                                                                
    "code": 401,                                                                                                                                                                                                                                                                                                                                           
    "message": "Access Denied"                                                                                                                                                                                                                                                                                                                             
  }                                                                                                                                                                                                                                                                                                                                                        
}                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                           
 ›   JSONHTTPError: Unauthorized                                                                                                                                                                                                                                                                                                                           
```

But that might be because I’m logged in with the wrong account.

itamargiv:
> ...
> But that might be because I’m logged in with the wrong account.

It was decided to remove the `deployCommit` script altogether, which essentially means this entire section could be removed from the README, as we don't use it in our day-to-day development workflows anyhow.

lucaswerkmeister:
Okay, let’s make a new pull request for removing it instead.

PR #169, opened 2022-05-02
Introduce Tracker service and track Lexeme creations

lucaswerkmeister:
The MwTrack type is based on the [MediaWiki core documentation][1] as well as the [statsv documentation][2].

[1]: https://doc.wikimedia.org/mediawiki-core/master/js/#!/api/mw-method-track
[2]: https://wikitech.wikimedia.org/wiki/Graphite#statsv

Bug: T303394

### DIFF
--- a/index.html
+++ b/index.html
@@ -43,6 +43,11 @@
 						return `https://www.wikidata.org/w/index.php?search=${search}&ns146=1`;
 					},
 				},
+				tracker: {
+					increment( name ) {
+						console.log( `TRACK: increment ${name}` );
+					},
+				},
 				wikiRouter: { goToTitle( title ) { alert( `Navigating to: ${title}` ); } },
 			};
 
--- a/src/@types/mediawiki.ts
+++ b/src/@types/mediawiki.ts
@@ -18,12 +18,15 @@ interface MwConfig {
 	get( key: string ): unknown;
 }
 
+export type MwTrack = ( topic: `${'timing' | 'counter'}.${string}`, data?: object | number | string ) => void;
+
 export type MwUtilGetUrl = ( pageName: string|null, params?: Record<string, unknown> ) => string;
 
 export interface MediaWiki {
 	config: MwConfig;
 	message: MwMessages;
 	Api: new( defaultOptions?: MwApiOptions ) => MwApi;
+	track: MwTrack;
 	util: {
 		getUrl: MwUtilGetUrl;
 	};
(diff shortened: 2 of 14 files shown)


lucaswerkmeister:
Hm, now I’m second-guessing myself… should the tracker really be a *plugin*, that we’ll use in components? Or should it just be a *service* used in the relevant store action?

lucaswerkmeister:
> Hm, now I’m second-guessing myself… should the tracker really be a _plugin_, that we’ll use in components? Or should it just be a _service_ used in the relevant store action?

I think I’m leaning towards the latter now, I don’t foresee any tracking that we’d want to do from components at the moment. (Even error tracking should probably take place in actions.) But then we need to find a place outside of `plugins/` for the part of this code that we still want to keep ^^

micgro42:
> But then we need to find a place outside of `plugins/` for the part of this code that we still want to keep ^^

Mh, it doesn't really belong into `data-access/`. Should we have another top-level directory just for this? Maybe something like `tracking/` or `analytics/`? Not sure what kind of naming we use in other places. I think I would not want `services/` because that is going to become a catch-all like `helpers/` or `util/`.

lucaswerkmeister:
Ah, I didn’t see that comment in time and put it in services now :see_no_evil:

lucaswerkmeister:
My thinking was that we could also consider moving the other `plugins/` into `services/`… at least some of the code in there is at best indirectly related to plugins, IMHO (e.g. `MwMessagesRepository`, an implementation of `MessagesRepository`, which is an interface that’s used by the `Messages` plugin).

micgro42:
> My thinking was that we could also consider moving the other `plugins/` into `services/`… at least some of the code in there is at best indirectly related to plugins, IMHO (e.g. `MwMessagesRepository`, an implementation of `MessagesRepository`, which is an interface that’s used by the `Messages` plugin).

My main concern is that this leads to more ambiguity for where something belongs. For example, the `LanguageCodesProvider` is also a service and so on.

But I see you point about the `MwMessagesRepository`. Though that directory could probably be slimmed down a bit.

micgro42 (commented):
The code looks good in principle. However, if I add a `console.log` in `MwTracker` just before the line that sends the data, then it still gets never called. I need to `await` the call to the tracker in the action to see it work. Does it behave the same for you?

micgro42 (approved):
On second look, maybe that is just a fluke. Let's merge it now and try it out on beta.

PR #171, opened 2022-05-05
Simplify ItemLookup code a bit

lucaswerkmeister:
As far as I can tell, debouncedSearchForItems doesn’t need to be a ref, and I didn’t find any explanation for this in the comments of #78 either. Let’s try simplifying it…


guergana (commented):
Hi @lucaswerkmeister, line 6, importing DebounceFunc is not used anymore. It should be removed.

guergana (approved):
Works well and the code is easier to read this way. Thanks!

micgro42 (approved):

lucaswerkmeister:
Apparently this was originally based on Wikit’s `EntityLookup`, where the `debouncedUpdateMenuItems` function is [set in a similar way](https://github.com/wikimedia/wikidata-query-builder/blob/2b777fc641beb859221a0b11530e33ec4dc0235e/src/components/EntityLookup.vue#L76-L83); this might have been because `data()` there can’t access the `props`, which doesn’t seem to be a problem here.

PR #172, opened 2022-05-05
Expand Readme sections on linting and testing

micgro42:
Bug: T302321

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ npm t
 ```
 #### Dev Entry point
 
-To run the development entry point:
+To run the development entry point also known as the dev server:
 ```sh
 npm run dev
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,29 +18,56 @@ This will also install some Git pre-commit hooks that check any modified files a
...
 ```
-You can visit the dev entry point at <http://localhost:3000/>.
-Changes to the source files will be applied automatically,
-so you can start this once and then leave it running in the background.
 
-For instructions to see the form in the context of the special page,
-see the [WikibaseLexeme readme](https://gerrit.wikimedia.org/g/mediawiki/extensions/WikibaseLexeme/+/master/README.md).
+#### Testing
+
+Several npm scripts are available to only run a subset of these, such as:
```
lucaswerkmeister, line 57: “Of these” should be rephrased now that it’s no longer right after the `npm t` command.

README.md
```diff
@@ -18,29 +18,56 @@ This will also install some Git pre-commit hooks that check any modified files a
...
+npm run dev
+```
+You can visit the dev entry point at <http://localhost:3000/>.
+Changes to the source files will be applied automatically,
+so you can start this once and then leave it running in the background.
+
+For instructions to see the form in the context of the special page,
+see the [WikibaseLexeme readme](https://gerrit.wikimedia.org/g/mediawiki/extensions/WikibaseLexeme/+/master/README.md).
+
 To run all tests and linters:
```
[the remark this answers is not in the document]
micgro42: True, but I don't want another level of subheadings with "linting and testing" and with subsubheadings for "linting" and "testing". But maybe I could rephrase this slightly and promote it to appear directly below "regular development" :thinking:

README.md
micgro42: true, will try to find a better way to say it

README.md
```diff
@@ -18,29 +18,55 @@ This will also install some Git pre-commit hooks that check any modified files a
...
 
-To run all tests and linters:
+TL;DR: All checks are run with:
 ```sh
 npm t
 ```
+#### Dev Entry point
 
-Several npm scripts are available to only run a subset of these, such as:
+To run the development entry point:
```
lucaswerkmeister, line 27: Would something like “(aka the dev server)” make sense here? (That’s the term we use in the cypress section, and I think it might help to mention it here too.)

README.md
```diff
@@ -18,29 +18,55 @@ This will also install some Git pre-commit hooks that check any modified files a
...
-so you can start this once and then leave it running in the background.
 
-For instructions to see the form in the context of the special page,
-see the [WikibaseLexeme readme](https://gerrit.wikimedia.org/g/mediawiki/extensions/WikibaseLexeme/+/master/README.md).
+#### Testing
+
+Several npm scripts are available to only run a subset of tests, such as:
+```sh
+npm run test-only # only tests without linters
+npm run test:unit # only unit tests
```
lucaswerkmeister, line 59: Maybe we can remove the comment on this line? It’s pretty redundant now that there’s a better explanation of unit tests below, I guess :shrug:

lucaswerkmeister (commented):
looks pretty good, two minor comments

README.md
micgro42: makes sense :+1:

README.md
micgro42: True, it is somewhat self-explanatory

lucaswerkmeister (approved):

PR #173, opened 2022-05-05
Use new display section of wbsearchentities response

lucaswerkmeister:
We don’t need to support old Wikibase versions, so now that the patches for T104344 have been merged for a few weeks, it should be fine to use the display field instead of the label and description fields. (Note that the display field is guaranteed to be present even when empty since wikimedia/Wikibase@4fec51a6b8 / change I2de40f965a), so one test case that was omitting it needs to be updated.)

### DIFF
--- a/src/data-access/DevItemSearcher.ts
+++ b/src/data-access/DevItemSearcher.ts
@@ -42,7 +42,7 @@ export default class DevItemSearcher implements ItemSearcher {
 			origin: '*',
 		} ) ).then( ( r ) => r.json() ) as WbSearchEntitiesResponse;
 
-		return processResponse( response, 'en' );
+		return processResponse( response );
 	}
 
 }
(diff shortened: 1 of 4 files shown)


lucaswerkmeister:
Noticed this random TODO while looking through the code for… some reason related to T298150, I think, I don’t remember exactly what it was.

src/data-access/DevItemSearcher.ts
```diff
@@ -42,7 +42,7 @@ export default class DevItemSearcher implements ItemSearcher {
 			origin: '*',
 		} ) ).then( ( r ) => r.json() ) as WbSearchEntitiesResponse;
 
-		return processResponse( response, 'en' );
```
chukarave, line 45: Why is this redundant now?

lucaswerkmeister: Because `processResponse()` no longer has a `defaultLanguageCode` parameter; it was only needed to provide a language for the label and description when it wasn’t included in the response, but now we always have the language in the response itself.

chukarave (approved):
looks good to me

PR #176, opened 2022-05-06
Add lexicalCategorySuggestions to Config and to the ItemLookup

chukarave:
also adds a manual list to index.html for the
dev entrypoint

Bug: T298150

PR #177, opened 2022-05-06
Change ItemLookup error prop from optional to nullable

lucaswerkmeister:
This seems to work better with eslint once we start using `withDefaults` (#176); with the previous code, `vue/require-default-prop` complains that “Prop 'error' requires default value to be set”.


chukarave (approved):
lgtm

PR #178, opened 2022-05-09
Show suggested lexical categories in lookup

lucaswerkmeister:
Show the matching suggested items as soon as possible, and prepend them to the search results, taking care not to repeat the items in the search results.

In this implementation, if the search results contain an item that is in the suggestions but was not matched by its label in the suggestions (e.g. an alias, or a label in another language), then that item will be ordered among the search results, not among the suggested items.

Bug: T298150

### DIFF
--- a/src/components/ItemLookup.vue
+++ b/src/components/ItemLookup.vue
@@ -3,6 +3,7 @@ import { ref, computed } from 'vue';
 import { SearchedItemOption } from '@/data-access/ItemSearcher';
 import WikitLookup from './WikitLookup';
 import debounce from 'lodash/debounce';
+import escapeRegExp from 'lodash/escapeRegExp';
 import { useMessages } from '@/plugins/MessagesPlugin/Messages';
 
 interface Props {
@@ -11,7 +12,6 @@ interface Props {
 	value: string | null;
 	searchForItems: ( searchTerm: string, offset?: number ) => Promise<SearchedItemOption[]>;
 	error: { type: 'error'|'warning'; message: string } | null;
-	// eslint-disable-next-line vue/no-unused-properties
 	itemSuggestions?: SearchedItemOption[];
 }
 const props = withDefaults( defineProps<Props>(), {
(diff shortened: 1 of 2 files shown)


src/components/ItemLookup.vue
```diff
@@ -56,8 +56,20 @@ const onOptionSelected = ( value: unknown ) => {
 };
 
 const debouncedSearchForItems = debounce( async ( debouncedInputValue: string ) => {
+	const regExp = new RegExp( `\\b${escapeRegExp( debouncedInputValue )}`, 'i' );
```
micgro42, line 59: I wonder if we really should debounce the filtering of the static list. But it probably doesn't hurt :thinking:

tests/unit/components/ItemLookup.test.ts
```diff
@@ -142,7 +142,98 @@ describe( 'ItemLookup', () => {
 				},
 			] );
 		} );
+
+		it( ':itemSuggestions - shows matching suggestions before search results', async () => {
```
micgro42, line 146: This test should be part of the `:props` describe blog above

micgro42 (dismissed):
looks good overall, some minor issue with the test

src/components/ItemLookup.vue
lucaswerkmeister: I was wondering that as well… I could imagine people asking for both (possibly depending on their reduced-motion preference?). I’d leave it like this for now (simpler code) and see if product or UX have anything to say. (Subjectively, it still feels fast enough to me.)

tests/unit/components/ItemLookup.test.ts
lucaswerkmeister: It is part of the block, isn’t it?
image

micgro42: You're right! It seems GitHub is using different indents for code that is shown initially in a diff and code that is initially folded away and only revealed on user click. That confused me because the `it(` calls didn't line up.

micgro42:
In that case, I think I'm fine with this PR, but will wait with approval till tomorrow per our new policy

tests/unit/components/ItemLookup.test.ts
lucaswerkmeister: Right, that’s annoyed me before… I think for initially shown code it respects your tab width preference (4 characters for me), but for extra context it always seems to use 8 characters?

micgro42 (approved):

PR #185, opened 2022-05-11
Don’t debounce item suggestions in ItemLookup

lucaswerkmeister:
Move the handling of suggested items out of `debouncedSearchForItems`, and instead make `suggestedOptions` computed based on `itemSuggestions` and `searchInput`. (That requires moving `searchInput` up, which I’m not fully happy with, to be honest.) Deduplicating the menu items, to avoid having suggested items show up again in the search results, then happens in the computed `menuItems`.

----

Possible further development of #184. I’m on the fence whether I prefer it or not.

### DIFF
--- a/src/components/ItemLookup.vue
+++ b/src/components/ItemLookup.vue
@@ -25,22 +25,13 @@ const emit = defineEmits( {
 	},
 } );
 
-interface WikitMenuItem {
-	label: string;
-	description: string;
-	tag?: string;
-}
-
-interface MonolingualOption extends WikitMenuItem {
-	value: string;
-}
-const searchSuggestions = ref( [] as MonolingualOption[] );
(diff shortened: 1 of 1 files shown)


micgro42:
Could you explain a bit more / add to the commit message _why_ this is needed?

lucaswerkmeister:
> Could you explain a bit more / add to the commit message _why_ this is needed?

There isn’t really a “why”, to be honest… it’s something that I realized was possible, but didn’t want to include in #184. It could be considered as something positive – I quite like how simple `debouncedSearchForItems` becomes – or not. :shrug:

guergana:
There is a failing test.

lucaswerkmeister:
Will be included in #184 instead.

PR #186, opened 2022-05-12
Turn store language and lexicalCategory into SearchedItemOption

lucaswerkmeister:
Change the value prop of the `ItemLookup`, `LanguageInput` and `LexicalCategoryInput` components from a string (item ID) to a full `SearchedItemOption` object, and also change the type in the store, so that we use these objects (with an id, but also a display section) throughout. This will let us populate the lookups from the URL parameters.

Note that ItemLookup doesn’t fully react to incoming changes to the value prop yet, that will come later.

Bug: T308118

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -11,6 +11,7 @@
  * otherwise it’s preferred to directly use mutations.
  */
 
+import { SearchedItemOption } from '@/data-access/ItemSearcher';
 import LangCodeRetriever from '@/data-access/LangCodeRetriever';
 import LanguageCodesProvider from '@/data-access/LanguageCodesProvider';
 import LexemeCreator from '@/data-access/LexemeCreator';
@@ -36,14 +37,17 @@ export default function createActions(
 ): RootActions {
 	return {
 		async [ CREATE_LEXEME ]( { state, commit }: RootContext ): Promise<string> {
+			if ( !state.language || !state.lexicalCategory ) {
+				throw new Error( 'No language or lexical category!' ); // TODO
+			}
 			commit( CLEAR_ERRORS );
 			try {
 				const spellingVariant = state.languageCodeFromLanguageItem || state.spellingVariant;
 				const lexemeId = await lexemeCreator.createLexeme(
 					state.lemma,
 					spellingVariant,
-					state.language,
-					state.lexicalCategory,
+					state.language.id,
+					state.lexicalCategory.id,
 				);
 				tracker.increment( 'wikibase.lexeme.special.NewLexeme.js.create' );
 				return lexemeId;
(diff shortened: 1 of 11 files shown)


lucaswerkmeister:
> Note that ItemLookup doesn’t fully react to incoming changes to the value prop yet, that will come later.

I expect this will basically be something like
```js
watchEffect( () => {
	if ( props.value ) {
		searchInput.value = props.value.display.label?.value ?? props.value.id;
	}
} );
```
(from f56431e1c8), but it should have some tests too, so I think it makes more sense to make that a separate PR.

micgro42:
Still have to have a look at the code itself (will do that tomorrow), but I think the title of this PR is missing a word. (unless "store" is meant to be read twice in order to safe space?^^)

micgro42:
> Would “Turn … into SearchedItemOption” work?

Yep, that works for me

src/store/actions.ts
```diff
@@ -36,14 +37,17 @@ export default function createActions(
 ): RootActions {
 	return {
 		async [ CREATE_LEXEME ]( { state, commit }: RootContext ): Promise<string> {
+			if ( !state.language || !state.lexicalCategory ) {
+				throw new Error( 'No language or lexical category!' ); // TODO
```
micgro42, line 41: What will happen when this is triggered? Because I'm not sure that T305854 will get done soon, given that it is in Milestone X.

src/store/mutations.ts
```diff
@@ -20,10 +21,10 @@ export default {
 	[ SET_LEMMA ]( state: RootState, lemma: string ): void {
 		state.lemma = lemma;
 	},
-	[ SET_LANGUAGE ]( state: RootState, language: string ): void {
+	[ SET_LANGUAGE ]( state: RootState, language: SearchedItemOption ): void {
```
micgro42, line 24: It could also be `null` here, right?

src/store/actions.ts
lucaswerkmeister: Nothing user-visible happens, because we don’t add a message to the store but the promise rejection is still caught.

I’m not sure what should happen. Previously, we would’ve tried to create a lexeme with the empty string as language or lexical category ID, which works fine in the dev entry point and produces an API error on the special page. Should we keep this behavior for now?

src/store/mutations.ts
lucaswerkmeister: Yes, you’re right.

src/store/actions.ts
micgro42: Code-wise, I like this solution more. Though right now, people get at least a meaningful error message. OTOH, the current version isn't good enough anyway, so I think we can just go forward with this. Maybe we can increase the priority of that inline error-messages task soonish

micgro42 (approved):

src/store/actions.ts
lucaswerkmeister: (GitHub showed me the PR approval but not this new comment… :rage:)

Okay, let’s leave it like this for the moment.

PR #188, opened 2022-05-13
Load state from params at startup

lucaswerkmeister:
Introduce a store action to load the state from a params object, and call it at initialization time.

---

TODO:
- [x] add support for this in the dev entry point (open question in Mattermost)
- [x] should `createAndMount()` await the dispatch (and become async itself) or not?
- [x] Cypress test (depends on dev entry point support)


micgro42:
> One thing that’s still unclear here is the interaction of the specified spelling variant with the language code statement of the language item. I suggest leaving this for another pull request.

Isn't that answered in https://phabricator.wikimedia.org/T298154#7902429 ?

src/main.ts
```diff
@@ -43,6 +48,10 @@ export default function createAndMount(
 	const store = initStore( { ...services, languageCodesProvider } );
 	app.use( store );
 
+	if ( config.params !== undefined ) {
+		store.dispatch( INIT_FROM_PARAMS, config.params ); // without await
```
micgro42, line 52: I wonder if we want to name it differently in this app, as "params" is a very generic term. Maybe `INIT_FROM_URL` would be more to the point of what is going on. Or maybe `INIT_WITH_DATA` or just `INIT` could also be plausible, as the app doesn't really know/care where that data is coming from?

Though I'm not sure about this yet, just a thought

lucaswerkmeister:
> > One thing that’s still unclear here is the interaction of the specified spelling variant with the language code statement of the language item. I suggest leaving this for another pull request.
> 
> Isn't that answered in https://phabricator.wikimedia.org/T298154#7902429 ?

Kind of, but I thought it might need enough code changes to implement that it would make more sense in a separate pull request. But looking at it again, I think we almost have everything we need (specifically, that `HANDLE_LANGUAGE_CHANGE` already sets the spelling variant to empty, avoiding the issue of a stale spelling variant surviving language changes)… I can try it out.

micgro42:
> > > One thing that’s still unclear here is the interaction of the specified spelling variant with the language code statement of the language item. I suggest leaving this for another pull request.
> > 
> > 
> > Isn't that answered in [phabricator.wikimedia.org/T298154#7902429](https://phabricator.wikimedia.org/T298154#7902429) ?
> 
> Kind of, but I thought it might need enough code changes to implement that it would make more sense in a separate pull request. But looking at it again, I think we almost have everything we need (specifically, that `HANDLE_LANGUAGE_CHANGE` already sets the spelling variant to empty, avoiding the issue of a stale spelling variant surviving language changes)… I can try it out.

I mean, I'm happy to review more smaller pull requests than a single big one. To me, it just sounded like the requirement for this situation was considered unclear in the first part of the sentence, and so I wondered if that was actually the case or whether the requirement was sufficiently defined by that comment on phabricator.

lucaswerkmeister:
> I mean, I'm happy to review more smaller pull requests than a single big one. To me, it just sounded like the requirement for this situation was considered unclear in the first part of the sentence, and so I wondered if that was actually the case or whether the requirement was sufficiently defined by that comment on phabricator.

Well, it was kind of both – I thought the extra code changes might require behavioral changes too (e.g. to show the spelling variant input in some more cases). But I think that’s no longer necessary.

src/store/actions.ts
```diff
@@ -68,13 +89,41 @@ export default function createActions(
...
+			if ( params.spellVarCode !== undefined ) {
+				commit( SET_SPELLING_VARIANT, params.spellVarCode );
+			}
+			if ( params.language !== undefined ) {
+				commit( SET_LANGUAGE, {
+					id: params.language.id,
+					display: params.language.display,
+					// don’t copy params.language.languageCode here
+				} );
+				await dispatch( HANDLE_ITEM_LANGUAGE_CODE, params.language.languageCode );
```
micgro42, line 112: This is not yet doing the right thing for situations where there is no language code in the URL. It currently sets the language code to "undefined" which in this app means that the language code request is still _loading_, but that doesn't actually happen.

So in this case here, if there is no language code, then it should be set to `null`.

You can try this out with the "language item without language code" link in the Readme

src/main.ts
lucaswerkmeister: I like “params” better than “data” because at least “params” is often used as the term for the URL params specifically… what if we rearrange the words a bit? We could use “init params” is a noun phrase (like in the `InitParams` interface already), and change the action name from “init from (generic) params” to something like “update from init params” or “handle init params”.

src/store/actions.ts
lucaswerkmeister: > So in this case here, if there is no language code, then it should be set to `null`.

Sounds good, let’s do that on the PHP side :P IMHO the current `string | undefined | null | false` type of the `languageCodeFromLanguageItem` is already confusing enough without introducing another type and associated translation logic like “undefined in the init params should become null in the state”.

micgro42: Though we can of course add that to the payload, but it feels redundant.

lucaswerkmeister:
> I mean, I'm happy to review more smaller pull requests than a single big one.

I put it into #189 because I think it’ll be good to merge with a separate commit message (whereas this PR can be squashed at the end).

src/store/actions.ts
lucaswerkmeister: (In that case, I’d change the type declaration here to make the `languageCode` non-optional within the `language`, as long as the `language` itself is present.)

lucaswerkmeister: > To me "the language code does not exist in the initialization data for the language Item" is true iff "The language code does not exist on the language Item" in the Database. (or is deprecated)

Okay, but that still means we’re missing the `false` case, i.e. the best-rank language code statement has unknown value. I think it would be best if we made the value sent by PHP exactly the same as that returned by `LangCodeRetriever`: it’s present iff a language item was specified at all, either `null` (no / novalue statement), `false` (somevalue statement), or a string (to be validated in JS).

src/main.ts
lucaswerkmeister: Did that now.

micgro42: Yes, I like that more, thank you!

src/store/actions.ts
micgro42: There, I think this patch should implement that: https://gerrit.wikimedia.org/r/c/mediawiki/extensions/WikibaseLexeme/+/791622

PR #189, opened 2022-05-13
Use spelling variant over language code from language item

lucaswerkmeister:
Usually, at most one of these will be set to a truthy value: as soon as a language is selected, the spelling variant is cleared, and the input with which the user can change it again is only shown once the language code from the language item is known to not exist. However, when we load parameters from the URL, it is possible to set both at the same time; in this case, the spelling variant should be used over the language code from the language item (assuming the user submits the form without changing the language – otherwise, the spelling variant will be cleared again).

A final fallback to '' is necessary because, unlike the spelling variant, the language code from language item may not be a string; falling back to the empty string matches the previous behavior.

Bug: T298154

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -63,7 +63,7 @@ export default function createActions(
 			}
 			commit( CLEAR_ERRORS );
 			try {
-				const spellingVariant = state.languageCodeFromLanguageItem || state.spellingVariant;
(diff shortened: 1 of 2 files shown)


micgro42 (approved):

micgro42 (approved):

lucaswerkmeister:
Oops, the task ID should’ve been the subtask T308118. Oh well.

PR #191, opened 2022-05-16
Update Messages documentation

lucaswerkmeister:
Resolving an old FIXME.


micgro42:
Looks good, but should we disable those eslint warnings?

micgro42 (approved):

PR #192, opened 2022-05-18
Show per field errors

micgro42:
Bug: T308660


src/store/actions.ts
```diff
@@ -50,25 +52,85 @@ export type InitParams = {
...
+			}
+			if ( state.language ) {
+				if ( !state.languageCodeFromLanguageItem && !state.spellingVariant ) {
+					commit(
+						ADD_PER_FIELD_ERROR,
+						{ field: 'spellingVariantErrors', error: { messageKey: 'wikibaselexeme-newlexeme-error-no-spelling-variant' } },
+					);
+				} else {
+					formData.validSpellingVariant = state.spellingVariant ||
+						state.languageCodeFromLanguageItem as string;
```
micgro42, line 104: Not sure if it is possible to avoid that type cast here

tests/integration/NewLexemeForm.test.ts
```diff
@@ -235,6 +235,10 @@ describe( 'NewLexemeForm', () => {
 			await wrapper.trigger( 'submit' );
 
 			expect( createLexeme ).toHaveBeenCalledWith( 'foo', 'de', 'Q123', 'Q456' );
+
+			await nextTick();
+			await nextTick();
```
micgro42, line 240: Not sure if that should also be `flushPromises()`

lucaswerkmeister: probably… if this is between expectations on `createLexeme` and `goToTitle`, it’s not really about DOM modification, right?

src/store/actions.ts
```diff
@@ -50,25 +52,85 @@ export type InitParams = {
...
 	lexemeCreator: LexemeCreator,
 	langCodeRetriever: LangCodeRetriever,
 	languageCodesProvider: LanguageCodesProvider,
 	tracker: Tracker,
 ): RootActions {
 	return {
-		async [ CREATE_LEXEME ]( { state, commit }: RootContext ): Promise<string> {
-			if ( !state.language || !state.lexicalCategory ) {
-				throw new Error( 'No language or lexical category!' ); // TODO
+		[ GET_VALID_INPUTS ]( { state, commit }: RootContext ): ValidLexemeData {
```
lucaswerkmeister, line 69: the name “get” sounds like it should be a getter to me (though I understand why it must be an action – it commits mutations)… can we find an alternative verb? maybe “assemble”?

src/store/actions.ts
lucaswerkmeister: Pushed a commit avoiding it, feel free to revert if you don’t like i.

src/store/actions.ts
micgro42: Done in that other PR that will be uploaded soon™

src/store/actions.ts
```diff
@@ -50,25 +52,86 @@ export type InitParams = {
...
+			const formData: Partial<ValidLexemeData> = {};
+
+			if ( !state.lemma ) {
+				commit(
+					ADD_PER_FIELD_ERROR,
+					{ field: 'lemmaErrors', error: { messageKey: 'wikibaselexeme-newlexeme-error-no-lemma' } },
```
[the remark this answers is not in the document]
micgro42: let's keep your names. I'm in the process of polishing my commits right now anyway and I'm adjusting the names as I go

micgro42:
This is being replaced by a couple of smaller PRs

PR #193, opened 2022-05-23
Make ItemLookup error prop optional

lucaswerkmeister:
Apparently this is necessary to avoid warnings in debug mode:

> ```
> [Vue warn]: Missing required prop: "error"
>   at <ItemLookup label="Lexical category" placeholder="The Lexeme's category, e.g. 'verb'" value=null  ... >
>   at <LexicalCategoryInput modelValue=null onUpdate:modelValue=fn >
>   at <NewLexemeForm>
>   at <App>
> ```

---

Follows-up #177, which was apparently not quite right.


micgro42:
Ah yes! Maybe there should be a rule that complains about having a default value for a prop that is not optional, to make it more explicit.

micgro42 (approved):

PR #196, opened 2022-05-23
Change submit button while creating Lexeme

lucaswerkmeister:
Change the button’s label, and disable it, while the Lexeme is being created. When it’s done, reenable the button and change the label back.

Note that the button may still appear to be enabled when it’s clicked with the mouse (instead of another way to submit the form, like pressing Enter in the lemma input); this requires wmde/wikit#584 to fix.

In the tests, after rejecting the lexemeCreator promise, we apparently need to wait for the next Vue tick six times. I don’t know why, but five isn’t enough. I’m sorry.

Bug: T304511

### DIFF
--- a/tests/integration/NewLexemeForm.test.ts
+++ b/tests/integration/NewLexemeForm.test.ts
@@ -1,7 +1,7 @@
 import LanguageCodesProvider from '@/data-access/LanguageCodesProvider';
 import { Config, ConfigKey } from '@/plugins/ConfigPlugin/Config';
 import { LanguageCodesProviderKey } from '@/plugins/LanguageCodesProviderPlugin/LanguageCodesProvider';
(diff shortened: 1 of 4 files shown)


tests/integration/NewLexemeForm.test.ts
```diff
@@ -286,6 +286,68 @@ describe( 'NewLexemeForm', () => {
 			expect( goToTitle ).toHaveBeenCalledWith( 'Special:EntityPage/L123' );
 		} );
 
+		it( 'disables button when submitting and reenables on error', async () => {
+			let reject = undefined as unknown as ( reason: any ) => void;
```
micgro42, line 290: I think we can use `unknown` in place of `any` here to make the warning go away

micgro42 (approved):

PR #199, opened 2022-05-30
Show error if Lemma is missing

micgro42:
This is the second PR split out of #192 and building on top of #198. It includes an entire vertical slice of the functionality

### DIFF
--- a/src/store/RootState.ts
+++ b/src/store/RootState.ts
@@ -5,6 +5,10 @@ export interface SubmitError {
 	message?: string;
 }
 
+interface PerFieldError {
+	messageKey: string;
+}
+
 export default interface RootState {
 	lemma: string;
 	language: SearchedItemOption | null;
@@ -12,4 +16,7 @@ export default interface RootState {
 	lexicalCategory: SearchedItemOption | null;
 	spellingVariant: string;
 	globalErrors: SubmitError[];
+	perFieldErrors: {
+		lemmaErrors: PerFieldError[];
+	};
 }
--- a/src/store/mutations.ts
+++ b/src/store/mutations.ts
@@ -15,6 +15,8 @@ export const SET_LEXICAL_CATEGORY = 'setLexicalCategory';
 export const SET_SPELLING_VARIANT = 'setSpellingVariant';
 export const SET_LANGUAGE_CODE_FROM_LANGUAGE_ITEM = 'setLanguageCodeFromLanguageItem';
 export const ADD_ERRORS = 'addErrors';
+export const ADD_PER_FIELD_ERROR = 'addPerFieldError';
+export const CLEAR_PER_FIELD_ERRORS = 'clearPerFieldErrors';
 export const CLEAR_ERRORS = 'clearErrors';
 
 export default {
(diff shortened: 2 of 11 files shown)


src/store/RootState.ts
```diff
@@ -5,11 +5,18 @@ export interface SubmitError {
...
+
 export default interface RootState {
 	lemma: string;
 	language: SearchedItemOption | null;
 	languageCodeFromLanguageItem: string | undefined | null | false;
 	lexicalCategory: SearchedItemOption | null;
 	spellingVariant: string;
 	globalErrors: SubmitError[];
+	perFieldErrors: {
+		lemmaErrors: PerFieldError[];
```
lucaswerkmeister, line 20: Do these errors get cleared anywhere?

micgro42: Huh, I was so enamored with _adding_ those errors, that I fully forgot about clearing them! Will add the functionality right away!

micgro42: Added it. Not sure how to best test it though :thinking:

lucaswerkmeister (commented):
I just realized that this is different from the plan of action in T305854#7937644: we originally planned to check the lemma length on `input` events on the input, not just when the submit button is clicked…

micgro42:
> I just realized that this is different from the plan of action in [T305854#7937644](https://phabricator.wikimedia.org/T305854#7937644)#7937644: we originally planned to check the lemma length on input events on the `input`, not just when the submit button is clicked…

But that comment explicitly discusses checking for too long a lemma and an empty lemma in two different sections, or am I misreading it?

lucaswerkmeister:
> > I just realized that this is different from the plan of action in [T305854#7937644](https://phabricator.wikimedia.org/T305854#7937644)#7937644: we originally planned to check the lemma length on input events on the `input`, not just when the submit button is clicked…
> 
> But that comment explicitly discusses checking for too long a lemma and an empty lemma in two different sections, or am I misreading it?

You’re right, I got confused and thought this PR was already checking the length.

Is the length check not implemented at all yet?

micgro42:
> Is the length check not implemented at all yet?

I think that part was not yet started to avoid conflicts with this work here.

src/store/mutations.ts
```diff
@@ -15,6 +15,8 @@ export const SET_LEXICAL_CATEGORY = 'setLexicalCategory';
 export const SET_SPELLING_VARIANT = 'setSpellingVariant';
 export const SET_LANGUAGE_CODE_FROM_LANGUAGE_ITEM = 'setLanguageCodeFromLanguageItem';
 export const ADD_ERRORS = 'addErrors';
+export const ADD_PER_FIELD_ERROR = 'addPerFieldError';
+export const CLEAR_PER_FIELD_ERROR = 'clearPerFieldError';
```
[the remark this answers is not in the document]
micgro42: Done.

src/store/RootState.ts
lucaswerkmeister: I guess that feels like an integration test? But we could also do that at the end of the chain, for all inputs at once (errors are added on submit and cleared one by one as input is added).

lucaswerkmeister (approved):

PR #200, opened 2022-05-30
Show error if lexical category is missing at submit

micgro42:
This targets #199 and has been split out of #192

### DIFF
--- a/src/components/LexicalCategoryInput.vue
+++ b/src/components/LexicalCategoryInput.vue
@@ -4,6 +4,8 @@ import { SearchedItemOption } from '@/data-access/ItemSearcher';
 import { useMessages } from '@/plugins/MessagesPlugin/Messages';
 import { useItemSearch } from '@/plugins/ItemSearchPlugin/ItemSearch';
 import { useConfig } from '@/plugins/ConfigPlugin/Config';
+import { useStore } from 'vuex';
+import { computed } from 'vue';
 
 interface Props {
 	modelValue: SearchedItemOption | null;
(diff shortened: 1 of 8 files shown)


src/components/LexicalCategoryInput.vue
```diff
@@ -19,6 +21,18 @@ const searchForItems = searcher.searchItems.bind( searcher );
...
+const error = computed( () => {
+	if ( !store.state.perFieldErrors.lexicalCategoryErrors.length ) {
+		return null;
+	}
+	return {
+		type: 'error' as const,
+		message: messages.getUnescaped(
+			store.state.perFieldErrors.lexicalCategoryErrors[ 0 ].messageKey,
+		),
+	};
```
lucaswerkmeister, line 34: I still think this code should be extracted into a function somewhere, taking the `PerFieldError[]` and returning an object `{ type, message }`. I guess it would be another plugin, so that it can access the `messages` without it being awkward? (Or take those as another parameter.)

src/components/NewLexemeForm.vue
```diff
@@ -51,6 +51,9 @@ const lexicalCategory = computed( {
 	},
 	set( newLexicalCategory: SearchedItemOption | null ): void {
 		store.commit( SET_LEXICAL_CATEGORY, newLexicalCategory );
+		if ( newLexicalCategory ) {
+			store.commit( CLEAR_PER_FIELD_ERRORS, 'lexicalCategoryErrors' );
```
lucaswerkmeister, line 55: This effectively hard-codes the knowledge that “no lexical category selected” is the only possible error for this field. That feels a bit awkward to me, but I can’t think of a great way to avoid it.

lucaswerkmeister (approved):

lucaswerkmeister (approved):

src/components/NewLexemeForm.vue
micgro42: yeah, I think after all these per-field-errors are done, it might be sensible to have another overall look and think about whether we can come up with a better solution. Though the result might be not to change it, because yagni as we don't have any other errors yet, might never have any other errors and if we have other errors whereever they might be coming from might determine the best way to set this up.

(though it will be interesting to look at the "lemma too long error")

src/components/LexicalCategoryInput.vue
micgro42: mh, maybe let's do this in an extra pull request at the end

PR #201, opened 2022-05-30
Show error if language is missing

micgro42:
Split out of #192

### DIFF
--- a/src/components/LanguageInput.vue
+++ b/src/components/LanguageInput.vue
@@ -23,6 +23,14 @@ const searchForItems = searcher.searchItems.bind( searcher );
 const store = useStore();
 
 const error = computed( () => {
+	if ( store.state.perFieldErrors.languageErrors.length ) {
+		return {
+			type: 'error' as const,
+			message: messages.getUnescaped(
+				store.state.perFieldErrors.languageErrors[ 0 ].messageKey,
+			),
+		};
+	}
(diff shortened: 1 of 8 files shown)


micgro42:
> Seems to be missing the clearing-errors part.

Argh! I did feel to me like it was surprisingly few changes when amending this commit. Should be fixed now, I think.

lucaswerkmeister:
> > Seems to be missing the clearing-errors part.
> 
> Argh! I did feel to me like it was surprisingly few changes when amending this commit. Should be fixed now, I think.

Hm, it’s still not working for me but I’m not sure why… I press Enter on an empty form, see all three errors, and then when I select a language item the error stays there (whereas for the lexical category it vanishes).

micgro42:
> > > Seems to be missing the clearing-errors part.
> > 
> > 
> > Argh! I did feel to me like it was surprisingly few changes when amending this commit. Should be fixed now, I think.
> 
> Hm, it’s still not working for me but I’m not sure why… I press Enter on an empty form, see all three errors, and then when I select a language item the error stays there (whereas for the lexical category it vanishes).

Because that setter is never called. We're dispatching an action in a different event handler instead. *sigh*

/me moves commit around and removes setter (maybe in a second PR)

lucaswerkmeister (approved):

PR #203, opened 2022-05-30
Clean up unused setter and simplify getter

micgro42:
That setter became unused in #130 / 003c8e7

### DIFF
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -16,7 +16,6 @@ import LexicalCategoryInput from '@/components/LexicalCategoryInput.vue';
 import ErrorMessage from '@/components/ErrorMessage.vue';
 import {
 	CLEAR_PER_FIELD_ERRORS,
-	SET_LANGUAGE,
 	SET_LEMMA,
 	SET_LEXICAL_CATEGORY,
 	SET_SPELLING_VARIANT,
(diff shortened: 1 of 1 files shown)


lucaswerkmeister (changes requested):
eslint points out the setter import is now unused (but I can’t add a comment on it, let alone a suggested change, because github)

lucaswerkmeister (approved):
Is there a reason why the `store.dispatch( HANDLE_LANGUAGE_CHANGE, newLanguage )` can’t be in the computed setter, so it’s still more consistent with the other fields? :thinking:

lucaswerkmeister:
> Is there a reason why the `store.dispatch( HANDLE_LANGUAGE_CHANGE, newLanguage )` can’t be in the computed setter, so it’s still more consistent with the other fields? thinking

Hm, that doesn’t seem to work, but I don’t know why…

micgro42:
> > Is there a reason why the `store.dispatch( HANDLE_LANGUAGE_CHANGE, newLanguage )` can’t be in the computed setter, so it’s still more consistent with the other fields? thinking
> 
> Hm, that doesn’t seem to work, but I don’t know why…

It seems to work for me and I like it more. I updated the PR :)

src/components/NewLexemeForm.vue
```diff
@@ -41,8 +40,11 @@ const language = computed( {
 	get(): SearchedItemOption | null {
 		return store.state.language;
 	},
-	set( newLanguage: SearchedItemOption | null ): void {
-		store.commit( SET_LANGUAGE, newLanguage );
+	async set( newLanguage: SearchedItemOption | null ): Promise<void> {
+		await store.dispatch( HANDLE_LANGUAGE_CHANGE, newLanguage );
+		if ( newLanguage ) {
+			store.commit( CLEAR_PER_FIELD_ERRORS, 'languageErrors' );
```
lucaswerkmeister, line 46: Hm, should this really wait until we’re done looking up the language code of the language item? (But can be switched around later.)

lucaswerkmeister (approved):
I think I just copied the wrong code or something ^^

src/components/NewLexemeForm.vue
micgro42: Yeah, let's think about this later

PR #204, opened 2022-05-30
Show error of missing spelling variant

micgro42:
Bug: T308660

### DIFF
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -64,8 +64,11 @@ const spellingVariant = computed( {
 	get(): string {
 		return store.state.spellingVariant;
 	},
-	set( newSpellingVariant: string ): void {
+	set( newSpellingVariant: string | null ): void {
 		store.commit( SET_SPELLING_VARIANT, newSpellingVariant );
+		if ( newSpellingVariant ) {
+			store.commit( CLEAR_PER_FIELD_ERRORS, 'spellingVariantErrors' );
+		}
 	},
 } );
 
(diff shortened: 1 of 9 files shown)


lucaswerkmeister (changes requested):
This seems to introduce a new bug, though I don’t yet see the error in the code. Steps to reproduce:

1. Select a language item without a language code, e.g. type “n” to get the suggestion “north”.
2. Enter the spelling variant en (English).
3. Click the “create lexeme” button. There are errors on the lemma and lexical category.
4. Clear the spelling variant input.
5. Click the “create lexeme” button.

The spelling variant now gets an error (empty, please make entry), but it also gets the value en (English) again. This doesn’t happen on `main`.

micgro42:
o.O

micgro42:
Found it. Somehow changing the error prop triggers a complete rerender of either the `WikitLookup` or Wikit's `Lookup` component from scratch. That has the sideffect that the text in the `<input>` is determined by the content of the `searchInput` in `SpellingVariantInput.vue`. However, that ref's value is only updated iff the new value is _not_ the empty string.

That was no problem back when clicking "submit" didn't cause a rerender of that component. But now it sometimes does and thus the error. I'll add a commit to this PR that fixes that.

lucaswerkmeister (approved):
Found some more inconvenient behavior: if the lexeme language has an item without a language code selected, then just tabbing through that field will wipe whatever you had in the spelling variant input (e.g. if you select language and spelling variant first, then tab back up to the lemma, now your spelling variant is gone). But that’s probably unrelated to this change.

PR #205, opened 2022-05-31
Validate Lemma input length

chukarave:
add messages to the Message plugin
add maxLemmaLength to the Config plugin
use to validate input length and display error inline

Bug: T308654

### DIFF
--- a/tests/unit/components/LemmaInput.test.ts
+++ b/tests/unit/components/LemmaInput.test.ts
@@ -14,7 +14,7 @@ describe( 'LemmaInput', () => {
 
 	let store: Store<RootState>;
 
-	function createComponent( config: Record<string, unknown> = {} ) {
+	function createComponent( config: Record<string, unknown> = { maxLemmaLength: 8 } ) {
 		store = initStore( {
 			lexemeCreator: unusedLexemeCreator,
 			langCodeRetriever: unusedLangCodeRetriever,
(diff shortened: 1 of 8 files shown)


tests/unit/components/LemmaInput.test.ts
```diff
@@ -57,6 +57,17 @@ describe( 'LemmaInput', () => {
 
 			expect( lemmaInputWrapper.get( '.wikit-ValidationMessage--error' ).text() ).toContain( '⧼wikibaselexeme-newlexeme-error-no-lemma⧽' );
 		} );
+
+		it( ':error - displays an error message in case input is too long', async () => {
+			const tooLongValue =
```
micgro42, line 62: This does indeed work for basically any value here, because you're not setting `config.maxLemmaLength` anywhere. Have a look at the `createComponent(...)` helper function in this class. It allows you to override part of the default configuration. You can do that with a smaller value than 1000, then it is easier to test.

Also, this should include a second test case that makes sure you're counting code-points, not double-bytes. That is, "`🏳️‍🌈`" should be counted as length 4, not length 6.

tests/unit/components/LemmaInput.test.ts
```diff
@@ -14,7 +14,7 @@ describe( 'LemmaInput', () => {
 
 	let store: Store<RootState>;
 
-	function createComponent( config: Record<string, unknown> = {} ) {
+	function createComponent( config: Record<string, unknown> = { maxLemmaLength: 8 } ) {
```
chukarave, line 17: setting it here doesn't seem to make a difference, but that could also mean the validation isn't actually working

micgro42 (approved):

PR #206, opened 2022-05-31
Upgrade dependencies

micgro42:
There seems to be an issue with `@vue/test-utils`. As soon as I update that, I get type errors during testing. This needs some more investigation and then probably an issue opened there. Also, I think I would like to hold of on upgrading to jest 28 until that has been investigated a bit more.

That being said, the upgrades in this PR are ready to be merged and should be uneventful. Would be great if can get this done today before dependabot opens a whole lot of PRs tomorrow.

The idea is to squash the commits in this PR when merging it.


micgro42:
Ugh. Apparently, there is a peer dependency conflict between `@vue/eslint-config-typescript` 10 and `eslint-plugin-vue` 9. I've downgraded the latter back to the latest 8.x version for now.

lucaswerkmeister (approved):

PR #214, opened 2022-06-01
Use wikibase-entityselector-notfound for “no results”

lucaswerkmeister:
We can use this existing Wikibase message for the “no results” copy of the item lookup and spelling variant input, just like we’re already using `wikibase-shortcopyrightwarning` elsewhere.

Bug: T304340

----

I’m not sure if I like using this message for the spelling variant lookup as well… the English version seems fine, but it’s possible that some translations of the message are actually phrased in a way that is more specific to an entity selector, and doesn’t work as well for a spelling variant input.

### DIFF
--- a/src/components/ItemLookup.vue
+++ b/src/components/ItemLookup.vue
@@ -150,7 +150,7 @@ const messages = useMessages();
 		@input="onWikitOptionSelected"
 	>
 		<template #no-results>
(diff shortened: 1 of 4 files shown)


guergana:
Looks good!

guergana (approved):
Looks good! Thanks!

PR #215, opened 2022-06-01
Don’t reset spelling variant when language stays the same

lucaswerkmeister:
When the user tabs through the form, leaving an item lookup will emit a “value” update for that lookup, even if the value didn’t change. That’s fine for the lexical category lookup, but problematic for the language lookup, because it would wipe whatever spelling variant the user had already entered, and also retry looking up the language code of the item. Instead, only have the `HANDLE_LANGUAGE_CHANGE` action reset the relevant state if the language actually changed, not if it stayed the same. (Note that we still commit the `SET_LANGUAGE` mutation, as the display section of the item may have changed.)

Bug: T201587

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -144,10 +144,15 @@ export default function createActions(
 			}
 		},
 		async [ HANDLE_LANGUAGE_CHANGE ](
-			{ commit, dispatch }: RootContext,
+			{ state, commit, dispatch }: RootContext,
 			newLanguageItem: SearchedItemOption | null,
 		): Promise<void> {
+			const oldLanguageItem = state.language;
 			commit( SET_LANGUAGE, newLanguageItem );
+			if ( oldLanguageItem?.id === newLanguageItem?.id ) {
+				return;
+			}
+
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
Alternatively, checking this in the `ItemLookup` before emitting the event also seems to work in practice:
```diff
diff --git a/src/components/ItemLookup.vue b/src/components/ItemLookup.vue
index a3c173536c..3159176171 100644
--- a/src/components/ItemLookup.vue
+++ b/src/components/ItemLookup.vue
@@ -67,6 +67,9 @@ const menuItems = computed( (): SearchedItemOption[] => {
 // the parent component tree has finished processing the `'update:modelValue'` event emitted here.
 const lastSelectedOption = ref( null as SearchedItemOption | null );
 const onOptionSelected = ( value: SearchedItemOption | null ) => {
+	if ( value === lastSelectedOption.value ) {
+		return;
+	}
 	lastSelectedOption.value = value;
 	emit( 'update:modelValue', value );
 };
```
But I think I prefer doing it in the store.

lucaswerkmeister:
We could also still add the other approach, I suppose ^^

PR #216, opened 2022-06-01
Reduce redundancy in integration test

micgro42:
Most tests need only some minor adjustments to the overall form setup.
That setup is much more obvious if it is the only thing handed to the
createForm function, overriding specific defaults instead of repeating
them.


chukarave (approved):
lgtm

PR #217, opened 2022-06-01
Add integration tests for clearing empty-input errors

micgro42:
Bug: T308660

### DIFF
--- a/tests/integration/NewLexemeForm.test.ts
+++ b/tests/integration/NewLexemeForm.test.ts
@@ -285,4 +285,155 @@ describe( 'NewLexemeForm', () => {
 
 	} );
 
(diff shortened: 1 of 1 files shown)


micgro42:
I wonder if there should be another refactoring step to turn 

```js
			const lexicalCategoryInput = wrapper.find( '.wbl-snl-lexical-category-lookup input' );
			await lexicalCategoryInput.setValue( '=Q456' );
			await wrapper.find( '.wbl-snl-lexical-category-lookup .wikit-OptionsMenu__item' ).trigger( 'click' );
```

into something like

```js
			await setLexicalCategoryValue();
```

with maybe an optional argument for the actual value?

chukarave (approved):
This seems to cover all relevant cases

PR #222, opened 2022-06-08
Test that max lemma length is passed into message

lucaswerkmeister:
Restructure createComponent() so that it’s possible to usefully override the `provide` globals as well, and use that to inject a custom messages plugin implementation.

Bug: T308654


chukarave (approved):
looks good to me

PR #223, opened 2022-06-10
export ItemLookup searchInput to be reachable from field components

chukarave:
Currently implemented only for Lexeme language and lexical category
(spelling variant implementation still pending).
for each field, add a searchInput store value, using which the
component receives the correct error message in the case input
is empty/invalid.

Bug: T310134

### DIFF
--- a/src/components/ItemLookup.vue
+++ b/src/components/ItemLookup.vue
@@ -2,7 +2,6 @@
 import {
 	computed,
 	ref,
-	watchEffect,
 } from 'vue';
 import { SearchedItemOption } from '@/data-access/ItemSearcher';
 import WikitLookup from './WikitLookup';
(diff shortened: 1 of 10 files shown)


src/components/ItemLookup.vue
```diff
@@ -15,18 +15,23 @@ interface Props {
...
 const emit = defineEmits( {
 	'update:modelValue': ( value: Props['value'] ) => {
 		return value === null || /^Q\d+$/.test( value.id );
 	},
+	'update:modelSearchInput': ( value: Props['searchInput'] ) => {
+		return true;
+	},
 } );
 
 const searchInput = ref( '' );
```
lucaswerkmeister, line 37: I only noticed just now that we have two `searchInput`s here now: the prop and the ref. Pretty sure we should combine these in some way. (And I think we can actually remove the `watchEffect` below this line, and instead have the store control the `searchInput`s in `HANDLE_INIT_PARAMS`.)

lucaswerkmeister:
I force-pushed a new version that squashes all the commits so far, rebases onto some PRs that were merged in the meantime, and takes out the different messages in the store (that’s #229 now). I think this might be ready to merge (with another PR for the spelling variant search input TBD).

lucaswerkmeister:
Spelling variant PR (also without actually using new messages): #231

lucaswerkmeister (approved):

PR #225, opened 2022-06-10
Remove emits validators

lucaswerkmeister:
These seem like useless complication to me. The `SpellingVariantInput` validator was even incorrect, I would argue – it would reject `null` as an invalid value, when in fact that’s the correct value when the user hasn’t selected anything or cleared the input.

----

Noticed while working on https://github.com/wmde/new-lexeme-special-page/pull/223.

### DIFF
--- a/src/components/ItemLookup.vue
+++ b/src/components/ItemLookup.vue
@@ -23,11 +23,7 @@ const props = withDefaults( defineProps<Props>(), {
 	itemSuggestions: () => [],
 } );
 
-const emit = defineEmits( {
-	'update:modelValue': ( value: Props['value'] ) => {
-		return value === null || /^Q\d+$/.test( value.id );
-	},
-} );
+const emit = defineEmits( [ 'update:modelValue' ] );
 
 const searchInput = ref( '' );
 watchEffect( () => {
--- a/src/components/SpellingVariantInput.vue
+++ b/src/components/SpellingVariantInput.vue
@@ -35,11 +35,7 @@ languageCodesProvider.getLanguages().forEach(
 
 const menuItems = ref( [] as WikitMenuItem[] );
 
-const emit = defineEmits( {
-	'update:modelValue': ( selectedLang: string | null ) => {
-		return selectedLang;
-	},
-} );
+const emit = defineEmits( [ 'update:modelValue' ] );
 
 const searchInput = ref( '' );
 const onSearchInput = ( inputValue: string ) => {


lucaswerkmeister:
> Though I would note that these validators (same as prop validators, prop types, etc.) are only a development aid. They are not present in the actual production build.

Honestly, I didn’t find it helpful during development ^^ in #223, we didn’t remember how to specify an `emits` event name without a validator when other events already have a validator, so we ended up writing a pointless `returns true` validator (later removed in d1a625dedfff106e4eb22df204f75c790e64b411). But I uploaded #227 as an alternative.

micgro42:
> > Though I would note that these validators (same as prop validators, prop types, etc.) are only a development aid. They are not present in the actual production build.
> 
> Honestly, I didn’t find it helpful during development ^^ in #223, we didn’t remember how to specify an `emits` event name without a validator when other events already have a validator, so we ended up writing a pointless `returns true` validator (later removed in [d1a625d](https://github.com/wmde/new-lexeme-special-page/commit/d1a625dedfff106e4eb22df204f75c790e64b411)). But I uploaded #227 as an alternative.

I think, I would like to avoid going back to just a plain listing of events. What I could see as a workable way forward could be to use only [types for events](https://vuejs.org/guide/typescript/composition-api.html#typing-component-emits) by default and include a validator only when the emitted payload is particularly complex. What do you think?

lucaswerkmeister:
> I think, I would like to avoid going back to just a plain listing of events. What I could see as a workable way forward could be to use only [types for events](https://vuejs.org/guide/typescript/composition-api.html#typing-component-emits) by default and include a validator only when the emitted payload is particularly complex. What do you think?

Oh nice, I didn’t know about types for events. If it can be combined with a validator function, that sounds great to me.

In the meantime, I think we can merge #227 and close this?

micgro42:
> > I think, I would like to avoid going back to just a plain listing of events. What I could see as a workable way forward could be to use only [types for events](https://vuejs.org/guide/typescript/composition-api.html#typing-component-emits) by default and include a validator only when the emitted payload is particularly complex. What do you think?
> 
> Oh nice, I didn’t know about types for events. If it can be combined with a validator function when necessary, that sounds great to me.
> 
> In the meantime, I think we can merge #227 and close this?

I approved #227, but I'm not sure that they can be combined. The linked docs state:

> using either runtime declaration OR type declaration:

(emphasis in the original). I would interpret this as meaning XOR.

lucaswerkmeister:
I was hoping that exclusivity was only between the list/array syntax and the type syntax, but that the type syntax could still be combined with the object syntax… I’ll try it out, I guess ^^

lucaswerkmeister:
Dangit, doesn’t work.

```typescript
export declare function defineEmits<EE extends string = string>(emitOptions: EE[]): EmitFn<EE[]>;

export declare function defineEmits<E extends EmitsOptions = EmitsOptions>(emitOptions: E): EmitFn<E>;

export declare function defineEmits<TypeEmit>(): TypeEmit;
```

Oh well. Then I guess we either use `EmitsOptions` if we want any validation functions, or `TypeEmit` if we’re okay with types only, but never `EE[]`?

PR #226, opened 2022-06-13
Extract setSearchInput() helper for ItemLookup unit tests

lucaswerkmeister:
This becomes useful when the ItemLookup starts exporting the searchInput as a prop (#223); IMHO it’s also similar to 495410a2e3 (#218).

Bug: T310134


chukarave (approved):
Looks good to me

PR #229, opened 2022-06-14
Distinguish between message type for lookup fields

chukarave:
for each lookup field, ascertain which type of message key
is required (field empty/entry invalid) and use that while committing
This is a chunk broken down from [PR #223](https://github.com/wmde/new-lexeme-special-page/pull/223)

Bug: T310134

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -16,6 +16,7 @@ import LangCodeRetriever from '@/data-access/LangCodeRetriever';
 import LanguageCodesProvider from '@/data-access/LanguageCodesProvider';
 import LexemeCreator from '@/data-access/LexemeCreator';
 import Tracker from '@/data-access/tracking/Tracker';
+import MessageKeys from '@/plugins/MessagesPlugin/MessageKeys';
 import {
 	ActionContext,
 	ActionTree,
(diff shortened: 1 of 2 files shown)


micgro42 (approved):

lucaswerkmeister:
Can we add a test for this as well?

micgro42:
yeah, just noticed that after I clicked approve

lucaswerkmeister (dismissed):
LGTM; I wonder if we should also add the search inputs (as empty string) to the other test for `null` language / lexical category?

```diff
diff --git a/tests/unit/store/actions.test.ts b/tests/unit/store/actions.test.ts
index 74acf355e4..13adf5d835 100644
--- a/tests/unit/store/actions.test.ts
+++ b/tests/unit/store/actions.test.ts
@@ -196,7 +196,9 @@ describe( CREATE_LEXEME, () => {
 				return {
 					lemma: 'example lemma',
 					language: null,
+					languageSearchInput: '',
 					lexicalCategory: null,
+					lexicalCategorySearchInput: '',
 					spellingVariant: 'en',
 				} as RootState;
 			},
```

lucaswerkmeister (approved):

PR #231, opened 2022-06-14
Make SpellingVariantInput searchInput a prop

lucaswerkmeister:
Similar to what #223 does for the ItemLookup searchInput, make the SpellingVariantInput searchInput a prop and wire it up with the store. This will let the store show a different error message when the input isn’t empty but hasn’t been selected (not implemented in this commit).

Also, make HANDLE_INIT_PARAMS set the search input when the spelling variant has been loaded from the URL. Previously, the search input was left empty in that case, making it appear as if the spelling variant hadn’t been loaded. With this change, the input is set to just the language code (rather than the language name and code, combined via an interface message) – this should be improved further, but I think that needs a bit more consideration.

Bug: T310134

---

You can test the HANDLE_INIT_PARAMS behavior with [these URL parameters](http://localhost:3000/?initParams={%22language%22%3A{%22id%22%3A%22Q1860%22%2C%22display%22%3A{%22label%22%3A{%22language%22%3A%22en%22%2C%22value%22%3A%22English%22}%2C%22description%22%3A{%22language%22%3A%22en%22%2C%22value%22%3A%22language%22}}%2C%22languageCode%22%3Anull},%22spellVarCode%22:%22en%22}) ([staging site](https://d3002abfe74c2432e92dcb465ca996082011--new-lexeme-special-page.netlify.app/?initParams={%22language%22%3A{%22id%22%3A%22Q1860%22%2C%22display%22%3A{%22label%22%3A{%22language%22%3A%22en%22%2C%22value%22%3A%22English%22}%2C%22description%22%3A{%22language%22%3A%22en%22%2C%22value%22%3A%22language%22}}%2C%22languageCode%22%3Anull},%22spellVarCode%22:%22en%22}), [this pull request](https://9a0d889257f2fa64e6e9c5b29ce710b66142--new-lexeme-special-page.netlify.app/?initParams={%22language%22%3A{%22id%22%3A%22Q1860%22%2C%22display%22%3A{%22label%22%3A{%22language%22%3A%22en%22%2C%22value%22%3A%22English%22}%2C%22description%22%3A{%22language%22%3A%22en%22%2C%22value%22%3A%22language%22}}%2C%22languageCode%22%3Anull},%22spellVarCode%22:%22en%22})).

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -29,6 +29,7 @@ import {
 	SET_LEMMA,
 	SET_LEXICAL_CATEGORY,
 	SET_SPELLING_VARIANT,
+	SET_SPELLING_VARIANT_SEARCH_INPUT,
 } from './mutations';
 import RootState from './RootState';
 
@@ -173,6 +174,7 @@ export default function createActions(
 			}
 			if ( params.spellVarCode !== undefined ) {
 				commit( SET_SPELLING_VARIANT, params.spellVarCode );
+				commit( SET_SPELLING_VARIANT_SEARCH_INPUT, params.spellVarCode );
 			}
 			if ( params.language !== undefined ) {
 				commit( SET_LANGUAGE, {
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -19,6 +19,7 @@ import {
 	SET_LEMMA,
 	SET_LEXICAL_CATEGORY,
 	SET_SPELLING_VARIANT,
+	SET_SPELLING_VARIANT_SEARCH_INPUT,
 } from '@/store/mutations';
 import { useWikiRouter } from '@/plugins/WikiRouterPlugin/WikiRouter';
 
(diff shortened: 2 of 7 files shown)


src/store/actions.ts
```diff
@@ -173,6 +174,7 @@ export default function createActions(
 			}
 			if ( params.spellVarCode !== undefined ) {
 				commit( SET_SPELLING_VARIANT, params.spellVarCode );
+				commit( SET_SPELLING_VARIANT_SEARCH_INPUT, params.spellVarCode );
```
micgro42, line 177: :thinking: Does this then in practice show "en" or "English (en)" in the SpellingVariant field?

micgro42 (commented):
Looks mostly good, but I'm not yet fully sure about the change to the HANDLE_INIT_PARAMS action

src/store/actions.ts
lucaswerkmeister: It only shows `en` (hence the note in the commit message). I’m thinking perhaps these search input manipulations should be done inside the component after all (back to `watchEffect()`?) – it doesn’t feel right for the store to use the language code provider and the message plugin to reconstruct the same search input value. But I think that can be done after this is merged.

micgro42: Mh, good point. But I agree that we best look at it in another PR

PR #235, opened 2022-06-20
Add help link to spelling variant input

micgro42:
Bug: T298146


src/components/SpellingVariantInput.vue
```diff
@@ -84,6 +85,7 @@ const error = computed( () => {
 		),
 	};
 } );
+const helpUrl = 'https://www.wikidata.org/wiki/Help:Monolingual_text_languages'; // FIXME: inject?
```
micgro42, line 88: This should probably at least be [https://www.wikidata.org/wiki/Special:MyLanguage/Help:Monolingual_text_languages](https://www.wikidata.org/wiki/Special:MyLanguage/Help:Monolingual_text_languages), right? Not sure if it is worth injecting this link. It doesn't look like something that some 3rd party would want to point to someplace else, but not sure.

Thoughts?

lucaswerkmeister: Agree about Special:MyLanguage; the process to get a language code added would be different on other wikis, though…

lucaswerkmeister (commented):
Hm, I thought the link should only appear when the user tries to enter an invalid language code, but it doesn’t look like that in Figma. I guess that was an older design.

src/components/SpellingVariantInput.vue
lucaswerkmeister: Perhaps we can ask about it in the daily tomorrow?

micgro42: I don't think that there will be a daily tomorrow, but I'll bring it up with product

micgro42: Product suggestion is to put it into a message instead :heavy_check_mark:

src/components/SpellingVariantInput.vue
```diff
@@ -112,5 +114,23 @@ export default {
 		<template #no-results>
 			{{ messages.getUnescaped( 'wikibase-entityselector-notfound' ) }}
 		</template>
+		<template #suffix>
+			<span class="wbl-snl-spelling-variant-lookup__help-link">
+				<wikit-link :href="helpUrl"> {{ messages.getUnescaped(
```
lucaswerkmeister, line 119: That space in `> {{` looks suspicious to me, is it intentional?

src/components/SpellingVariantInput.vue
```diff
@@ -84,6 +85,7 @@ const error = computed( () => {
 		),
 	};
 } );
+const helpUrl = messages.get( 'wikibaselexeme-newlexeme-lemma-language-help-link-target' );
```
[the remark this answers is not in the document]
micgro42: Good idea, done.

lucaswerkmeister (approved):

PR #236, opened 2022-06-21
Fix margin-inline-end etc. not being transpiled

micgro42:


micgro42:
Taking inspiration from https://github.com/wmde/wikit/blob/next/vue-components/postcss.config.js, but wikit doesn't seem to have `autoprefixer` anywhere in its dependencies? o.O

micgro42:
I think this is ready to be reviewed now, I just wonder if we now need a .browserlistrc to define which browsers we want to support so that these postcss plugins know what they need to do?

lucaswerkmeister:
Hm, I don’t think this interacts too well with the transpiled+preserved Wikit styles… I think this now adds another copy of physical styles on e.g. `.wikit-Button` :/

Also, it uses `:dir()`, which is only supported in Firefox. (It’s not new – Firefox has supported it for half a decade – the other browsers are just asleep at the wheel there.) We need to add `postcss-dir-pseudo-class`, like in Wikit, I think.

micgro42:
Not sure how to tell postscript only to transform the CSS coming from `src/` directly and not any of the dependencies, for example wikit.

micgro42:
Closing, as this seems not feasible and the alternative #237 has been merged.

PR #237, opened 2022-06-23
Use backward compatible CSS with literal directions

micgro42:
This PR is effectively an alternative to #236 

We would have to transform the logical directions in order to support all older browser that we want to support. However, that transformation would also apply to wikit, effectively duplicating itself.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -4,7 +4,7 @@
 		"stylelint-config-standard-scss",
 		"stylelint-config-recommended-vue/scss"
 	],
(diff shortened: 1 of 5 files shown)


lucaswerkmeister (changes requested):
`NewLexemeForm.vue` has `margin-block-start`, we should also change that.

A stylelint rule to *enforce* physical properties might be useful, but I didn’t find one with a quick search.

lucaswerkmeister (approved):

PR #254, opened 2022-07-28
Show warning for anonymous users

lucaswerkmeister:
Using the standard Wikibase message, but without passing an argument into it. Special:NewLexeme passes the wikibase-entity-lexeme message into it (the argument used to be used as a sort of “your IP address will be visible in the history of the item/property” thing), but that message doesn’t actually exist; the $1 argument is also only used in one remaining translation (vec/Venetian), where I’ve asked the translator to update the translation so it doesn’t use $1 anymore. Once that’s done, I expect we’ll update Wikibase and WikibaseLexeme to also drop this message parameter.

On the dev entry point, the warning can be seen by adding ?isAnonymous=true to the URL.

Bug: T313466

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,6 +19,7 @@
 
 				const config = {
 					rootSelector: '#app',
+					isAnonymous: false,
 					licenseUrl: 'https://creativecommons.org/publicdomain/zero/1.0/',
 					licenseName: 'Creative Commons CC0',
 					wikibaseLexemeTermLanguages: new Map([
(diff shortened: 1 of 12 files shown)


micgro42:
Looking at the [dev entry point](https://dd4afc00dbda5213bdadb7d8f8799b05bbec--new-lexeme-special-page.netlify.app/?isAnonymous=true) made me wonder if just another intro-paragraph is really the way to go. I raised it in the phab ticket: https://phabricator.wikimedia.org/T313466#8115909

lucaswerkmeister:
Hm, CI reminded me to update the snapshot in the test – I force-pushed an update to that, but I’m not sure if snapshot testing still makes sense when we’re using the Wikit component…

micgro42:
That message uses links - does it come with them already prefilled and pointing to the right place?

micgro42 (approved):
> That message uses links - does it come with them already prefilled and pointing to the right place?

Mh, they do, though do not include a "get back to this page" functionality. Oh well, I guess that is part of a bigger (and different) story.

PR #256, opened 2022-07-29
Add asterisk to required form inputs

lucaswerkmeister:
The asterisk distances itself from anything else in the suffix slot using the same dimension as the gap between the label and the suffix as a whole; this is needed for the spelling variant input, where the suffix slot also contains a help link.

The asterisk is `aria-hidden` because we already mark the inputs themselves as `aria-required` and people don’t need two indications that the inputs are required.

Bug: T313892


micgro42 (approved):
That margin could also live in the component that uses the RequiredAsterisk, but I think the current solution is just as good.

PR #262, opened 2022-08-09
Install new version of wikit v.3.0.0-alpha.8

guergana:
Bug: [T312633](https://phabricator.wikimedia.org/T312633)


micgro42 (approved):
I can see the new styles in the inspector for both the TextInput and the Lookups and also notice the difference visually

PR #263, opened 2022-08-12
Style adjustments

chukarave:
The linter rule does not allow using 'rem' as a unit,
i've disabled it.
font-size for the copyright message should be 13px, the
rem in this case is relativly calculated from 1rem which is 16px.
margins between elements adapted to represent 16px as well.
The main content wrapper will get a top margin of 0.5rem
so that the gap between search-exiting or anonymous edit message
and the main form always remains 1.5rem.

Bug: T312652

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -12,6 +12,7 @@
 			"namespaces": ["wbl-snl-"]
 		},
 		"indentation": "tab",
-		"selector-max-id": 1
+		"selector-max-id": 1,
+		"unit-disallowed-list": null
 	}
 }
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -213,6 +213,7 @@ export default {
 
 	font-size: 0.8125rem;
 	font-style: italic;
+	margin-bottom: 0;
 }
 
 </style>
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -211,6 +211,7 @@ export default {
 .wbl-snl-copyright {
 	@include small-text;
 
+	font-size: 0.8125rem;
 	font-style: italic;
 }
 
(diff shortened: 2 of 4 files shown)


sai-san:
All looking great! Thanks Noa 🙏  I just noticed a small thing that wasn't explicitly mentioned in the ticket, but would be great if we could correct: the copyright message paragraph should be spaced like the rest of the elements in the form and the info panel (with 1rem of margin around it too). I think its margin is 1em at the moment, which results in different spacing.

Screenshot 2022-08-16 at 11 24 27

sai-san:
✅  The font-size of the copyright message is at least 13px
✅  The elements within the form have a default spacing of 1rem (generally 16px) between them**
❌  I'm not sure that the main content wrapper is displaying a margin of 0.5rem. The gap between search-exiting or anonymous edit message and the main form seems to be 1rem at the moment.

**I noticed that the copyright message still has a margin of 1em applied to it. This doesn't cause any problems because it overlaps with the 1rem margin applied to all elements. It might be a good idea to remove that just for consistency.

Screenshot 2022-08-16 at 11 24 27

chukarave:
@sai-san 
- the main content wrapper is not displaying the 0.5 margin here, it's part of the Gerrit patch as it's set from outside the Vue application. 
- Do you mean setting the copyright message margin explicitly to rem or removing it altogether?

sai-san:
Answering inline, @chukarave.

> * the main content wrapper is not displaying the 0.5 margin here, it's part of the Gerrit patch as it's set from outside the Vue application.

Got it! Is there a way I can check the Gerrit patch? I'm not sure where to find a Netlify link.

> * Do you mean setting the copyright message margin explicitly to rem or removing it altogether?

I think removing it altogether makes sense, since spacing is being defined at a different level. This shouldn't have any impact on the layout.

chukarave:
@sai-san 
> Is there a way I can check the Gerrit patch? I'm not sure where to find a Netlify link.
 
 as we discussed, this is not part of the Vue app so will be visible on Beta once merged

sai-san (approved):
All looking great! Thanks, @chukarave 💯

PR #266, opened 2022-08-23
Fix anon edit warning message styling

chukarave:
they were in the wrong component!

Bug: T312652

### DIFF
--- a/src/components/AnonymousEditWarning.vue
+++ b/src/components/AnonymousEditWarning.vue
@@ -25,8 +25,6 @@ const config = useConfig();
 @import "@wmde/wikit-tokens/variables";
 
 .wbl-snl-anonymous-edit-warning {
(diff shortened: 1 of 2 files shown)


src/components/AnonymousEditWarning.vue
```diff
@@ -26,5 +26,7 @@ const config = useConfig();
 
 .wbl-snl-anonymous-edit-warning {
 	margin: $dimension-spacing-small 0;
+	margin-top: 1rem;
```
micgro42, line 29: no, these lines are now just overriding what is in the line above

micgro42 (commented):
Looks good thanks!
Now 
```scss
```
can be deleted, I think, and then this is ready to go :)

micgro42 (approved):

PR #277, opened 2022-08-30
Update vue

guergana:
Bug: [T315964](https://phabricator.wikimedia.org/T315964)


micgro42 (approved):
I wonder if we should pin the vue version to what we're using in mediawiki? But that's independent of this PR, so this is good to go.

PR #308, opened 2022-10-19
Document chore procedure in README

lucaswerkmeister:
Partially based on the task descriptions of T313744 and T311696.

Bug: T311696

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ The following dependencies should be ignored:
 - Vue and Vuex:
   in production, we use the versions shipped by MediaWiki core,
   so we should use the same versions for testing.
+  The current versions shipped by MediaWiki core are listed in [foreign-resources.yaml](https://gerrit.wikimedia.org/g/mediawiki/core/+/master/resources/lib/foreign-resources.yaml).
 
 All other dependencies should generally be updated to the latest version.
 If you discover that a dependency should not be updated for some reason, please add it to the above list.
--- a/README.md
+++ b/README.md
@@ -88,3 +88,33 @@ After you've started the dev server, you can open the interactive browser tests 
 ```sh
 npm run cypress:open
 ```
+
+### Chore: Dependency Updates
+
+You can see which dependencies have new releases by first making sure your local dependencies are up-to-date by executing `npm ci` and then running `npm outdated`.
+The following dependencies should be ignored:
+
+- Wikit (i.e. `@wmde/wikit-tokens` and `@wmde/wikit-vue-components`):
+  we’re using a newer pre-release version and don’t want to downgrade to the latest full release.
+- Vue and Vuex:
+  in production, we use the versions shipped by MediaWiki core,
+  so we should use the same versions for testing.
+
+All other dependencies should generally be updated to the latest version.
+If you discover that a dependency cannot be updated for some reason, please add it to the above list.
+If a dependency can only be updated with substantial manual work,
+you can create a new task for it and skip it in the context of the current chore.
+
+The recommended way to update dependencies is to collect related dependency updates into grouped commit.
+This can be one commit for each of:
+
+- all ESLint-related dependency updates
+- all Jest-related dependency updates
+- all Vue-related dependency updates
+- all PostCSS/Stylelint-related dependency updates
+- `npm update` for all other dependency updates
+
+Make sure that all checks still pass for every commit.
+
+Once all dependency updates are merged, update the submodule version in WikibaseLexeme by running `npm run bump-special-new-lexeme` there:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,3 +88,33 @@ After you've started the dev server, you can open the interactive browser tests
...
+
+### Chore: Dependency Updates
+
+You can see which dependencies have new releases by first making sure your local dependencies are up-to-date by executing `npm ci` and then running `npm outdated`.
+The following dependencies should be ignored:
+
+- Wikit (i.e. `@wmde/wikit-tokens` and `@wmde/wikit-vue-components`):
+  we’re using a newer pre-release version and don’t want to downgrade to the latest full release.
+- Vue and Vuex:
+  in production, we use the versions shipped by MediaWiki core,
```
micgro42, line 100: Do we have a place that ought to show what version is shipped by MediaWiki core? If so, we may want to link that here

README.md
```diff
@@ -88,3 +88,33 @@ After you've started the dev server, you can open the interactive browser tests
...
+  so we should use the same versions for testing.
+
+All other dependencies should generally be updated to the latest version.
+If you discover that a dependency cannot be updated for some reason, please add it to the above list.
```
micgro42, line 104: I think it would be better to emphasize that this is a judgement call, not a limitation:

```suggestion
If you discover that a dependency should not be updated for some reason, please add it to the above list.
```

README.md
```diff
@@ -88,3 +88,33 @@ After you've started the dev server, you can open the interactive browser tests
...
+If a dependency can only be updated with substantial manual work,
+you can create a new task for it and skip it in the context of the current chore.
+
+The recommended way to update dependencies is to collect related dependency updates into grouped commit.
```
micgro42, line 108: typo:

```suggestion
The recommended way to update dependencies is to collect related dependency updates into grouped commits.
```

README.md
```diff
@@ -88,3 +88,33 @@ After you've started the dev server, you can open the interactive browser tests
...
+This can be one commit for each of:
```
micgro42, line 109: I would like to weaken this list a bit, as many variations of that might make sense.

```suggestion
For example, this can be one commit for each of:
```

README.md
micgro42: Also, a sentence outlining _why_ this is the recommended way would be good. Mentioning probably at least reviewability and debuggability. Though I don't have a good suggestion right now, this could also happen in the next iteration on this section.

README.md
guergana, line 100: Interesting, didn't know this, but it makes sense ofc.

README.md
```diff
@@ -88,3 +88,33 @@ After you've started the dev server, you can open the interactive browser tests
...
+
+- all ESLint-related dependency updates
+- all Jest-related dependency updates
+- all Vue-related dependency updates
+- all PostCSS/Stylelint-related dependency updates
+- `npm update` for all other dependency updates
+
+Make sure that all checks still pass for every commit.
+
+Once all dependency updates are merged, update the submodule version in WikibaseLexeme by running `npm run bump-special-new-lexeme` there:
```
guergana, line 119: Nice add.

guergana (approved):
It looks complete to me. Dunno if anyone else would think otherwise but I approve in the meantime,

README.md
guergana: +1 Sounds like a good idea!!

lucaswerkmeister: We could link to [foreign-resources.yaml](https://gerrit.wikimedia.org/g/mediawiki/core/+/master/resources/lib/foreign-resources.yaml), but I’m not sure how helpful it is (might be a bit long / confusing)?

README.md
lucaswerkmeister: Well, it was “for example” in the task description I based it on, but then I saw that the previous chore seemed to have followed this pattern fairly closely, so it felt strange to only call it an example… but I think that might sound better with the other change (to explain why this is the recommended way), so okay.

README.md
lucaswerkmeister: Done

README.md
micgro42: Maybe with something like "The current core version is listed in [foreign-resources.yaml](https://gerrit.wikimedia.org/g/mediawiki/core/+/master/resources/lib/foreign-resources.yaml)."
Then it is implied that one has to look through a list of things.

README.md
micgro42: The added sentences read like a good start. Thanks!

README.md
lucaswerkmeister: Done, thanks!

hasanakg (approved):

micgro42 (approved):
Looks good to me as well :+1:

PR #324, opened 2022-11-04
Remove GITHUB_TOKEN from comment-pull-request action

lucaswerkmeister:
This is no longer needed (it grabs the token on its own).


lucaswerkmeister:
Let’s see if it works and still leaves the comment here.

lucaswerkmeister:
Yay :tada:

lucaswerkmeister:
Nope, nevermind, that comment doesn’t prove anything – it seems to have been using the original version of the action:

image

I guess to verify this PR, we need to merge it, and then open a new test PR and possibly revert this change if it doesn’t work?

micgro42 (approved):
YOLO

lucaswerkmeister:
still works :tada: https://github.com/wmde/new-lexeme-special-page/pull/326#issuecomment-1305470868

PR #328, opened 2022-11-16
Move inputs and Create-Lexeme button to the front of the taborder

micgro42:
This is done with the main intention of effectively skipping the "terms of use" and "license" links when tabbing from the inputs to the create button, without making those links wholly inaccessible by tabbing.

TODO:
- [x] This needs a new pre-release of wikit first, see: wmde/wikit#644
- [x] sync with UX which version to prefer and why, the alternative is #329 

Bug: T322686

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,8 @@
 			"license": "GPL-2.0-or-later",
 			"dependencies": {
 				"@wmde/wikibase-datamodel-types": "^0.2.0",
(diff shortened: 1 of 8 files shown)


micgro42:
Closing this per the discussion in [⚓ T322686 tab behaviour issue on Special:NewLexeme](https://phabricator.wikimedia.org/T322686). We can reopen this if we decide we want it after all.

PR #329, opened 2022-11-16
Ignore the license links when tabbing

micgro42:
This is done with the main intention of effectively skipping the "terms of use" and "license" links when tabbing from the inputs to the create button.
They can still be clicked with the mouse, and remain accessible via link-navigation in screen readers (tested with VoiceOver, Orca, NVDA, and JAWS).

TODO:
- [ ] sync with UX which version to prefer and why, the alternative is #328

Bug: T322686

### DIFF
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -4,6 +4,7 @@ import { CREATE_LEXEME, HANDLE_LANGUAGE_CHANGE } from '@/store/actions';
 import {
 	computed,
 	ref,
+	onMounted,
 } from 'vue';
 import { useStore } from 'vuex';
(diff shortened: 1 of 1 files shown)


micgro42:
> > They can still be clicked with the mouse.
> 
> I'd also add "and remain accessible via keyboard by using control keys in screen readers, such us Voice Over" (just tested this again using the Netlify link). Again, it'd be great to validate if other screen reader technologies also make navigating to said links still possible.

I checked NVDA and JAWS in a Windows 10 VirtualBox, and it works there as well.

micgro42:
Closing this per the discussion in [⚓ T322686 tab behaviour issue on Special:NewLexeme](https://phabricator.wikimedia.org/T322686). We can reopen this if we decide we want it after all.

PR #331, opened 2022-11-20
Do not synthesize italic font in ToS&License copy

bgo-eiu:
This PR removes the italic styling from the license string because it makes it unreadable in non-Latin scripts. If the characters and font do not include italic variants, broken glyphs get rendered, which are particularly difficult to read in cursive scripts like those with Perso-Arabic alphabets. It is especially not a good idea here because the purpose of the NewLexeme page is explicitly multi-lingual.

In some contexts it may make sense to have language specific styles for emphasising text. However, this same string is not italicised elsewhere on Wikidata including the new item page so it is more consistent to not use an emphasised style here.

### DIFF
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -210,7 +210,8 @@ export default {
 
 .wbl-snl-copyright {
 	@include small-text;
-
+        font-style: italic;
+        font-synthesis: none;
 	font-size: 0.8125rem;
 	margin-bottom: 0;
 }
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -210,9 +210,10 @@ export default {
 
 .wbl-snl-copyright {
 	@include small-text;
-        font-style: italic;
-        font-synthesis: none;
+
 	font-size: 0.8125rem;
+	font-style: italic;
+	font-synthesis: none;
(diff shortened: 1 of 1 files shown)


micgro42:
@bgo-eiu Thank you for bringing this up!

As an alternative, instead of removing the `font-style: italics;`, we could also adjust [`font-synthesis`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-synthesis). That would mean that it still would be in italics for fonts that have dedicated glyphs for that, but it would be "roman" for fonts that do not.

What do you think?

bgo-eiu:
That would work as well - I actually tested this out in my user CSS and it worked smoothly. It has better browser support than I had realised. I will update the PR.

micgro42:
Thanks! I think it needs a bit of adjusting to the whitespace so that the linting is happy. I can do that, and then I'll merge it.

micgro42 (approved):

micgro42:
Mh, CI should not fail on this PR. I will have to look into that tomorrow.

micgro42:
I temporarily disabled that we require the code-health check to pass and merged it. I verified that it is fine locally.

Given that CI runs just fine for #332, I assume that the problem is that this request came from a fork of this repo? Not sure what the best way to handle this in the future is. Ideally, these actions should just run as usual after we approved them.

PR #332, opened 2022-11-24
Update Vite to v3 (stalled by WMF CI node version)

micgro42:
This is mainly here to test if GitHub CI works as expected due to the troubles with it in #331

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
 				"@types/lodash": "^4.14.187",
 				"@typescript-eslint/eslint-plugin": "^5.42.0",
 				"@typescript-eslint/parser": "^5.42.0",
-				"@vitejs/plugin-vue": "^2.3.4",
+				"@vitejs/plugin-vue": "^3.2.0",
 				"@vue/eslint-config-typescript": "^11.0.2",
 				"@vue/test-utils": "v2.2.1",
 				"@vue/vue3-jest": "^29.1.1",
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
I don’t think we can use Vite 3 yet – Wikimedia CI still uses Node 14.17.5 ([random recent Node build](https://integration.wikimedia.org/ci/job/mwgate-node14-docker/86567/console)), and Vite 3 requires 14.18 or later.

micgro42:
> I don’t think we can use Vite 3 yet – Wikimedia CI still uses Node 14.17.5 ([random recent Node build](https://integration.wikimedia.org/ci/job/mwgate-node14-docker/86567/console)), and Vite 3 requires 14.18 or later.

Ah, and we build it in CI to verify that it is indeed built correctly, right? Meh. Then I'll change this to draft until the WMDE infrastructure updates.

And I'll make a PR for the README, so that it actually says to ignore Vite 3.

lucaswerkmeister:
> Ah, and we build it in CI to verify that it is indeed built correctly, right? Meh.

Yeah, exactly.

> Then I'll change this to draft until the WMDE infrastructure updates.

(Which is T314470.)

> And I'll make a PR for the README, so that it actually says to ignore Vite 3.

Sounds good, thanks – apparently I totally forgot to include this in #308.

micgro42:
This will have to wait until WMF CI updates. README has been updated in #333

40N30E722296800Z:
Vite is upgraded to version 3 now.

lucaswerkmeister:
(Done in #445.)

PR #355, opened 2022-12-02
Update four patch versions

lucaswerkmeister:
```bash
npm i {@types/lodash,@vue/test-utils,husky,lint-staged}@latest
```

Bug: T323899


lucaswerkmeister:
The last dregs from `npm outdated`. There’s nothing that ties these four together, other than the fact that the version component being updated is the patch version.

package.json
```diff
@@ -54,12 +54,12 @@
...
 		"@rushstack/eslint-patch": "^1.2.0",
 		"@types/jest": "^29.2.3",
-		"@types/lodash": "^4.14.187",
+		"@types/lodash": "^4.14.191",
 		"@typescript-eslint/eslint-plugin": "^5.45.0",
 		"@typescript-eslint/parser": "^5.45.0",
 		"@vitejs/plugin-vue": "^2.3.4",
 		"@vue/eslint-config-typescript": "^11.0.2",
-		"@vue/test-utils": "v2.2.1",
+		"@vue/test-utils": "^2.2.5",
```
lucaswerkmeister, line 62: Oh neat, a mirrored caret operator!

micgro42: it's fascinating that this still worked. I guess it was then effectively pinned to that particular version?

micgro42 (approved):

package.json
lucaswerkmeister: Looks like it, yeah – on https://semver.npmjs.com/ a leading “v” seems to be accepted and ignored.

PR #386, opened 2023-01-17
Update axe-core to v4.6.2

lucaswerkmeister:
With Wikit v3.0.0-alpha.11, this should work now.

Bug: T326681


guergana (approved):
Tests seem to be passing so LGTM!

PR #452, opened 2023-05-23
Update dependencies

40N30E722296800Z:
Other dependencies

Bug: [T335735](T335735)


guergana (commented):
Hi, I have a nitpicky comment: Unrelated to the change but [the link to the pull request in the comment](https://github.com/wmde/new-lexeme-special-page/pull/452#issue-1722253404) is not correct, it should be linking to phabricator and it's somehow trying to find the ticket in github.

guergana (commented):
I don't think it's a big issue and doesn't mean we can not merge the ticket but if we follow the [instructions](https://github.com/wmde/new-lexeme-special-page#chore-dependency-updates) for updating dependencies we should group related dependencies. `axe-core`, `cypress` and `prettier` are not related dependencies and should ideally be updated in separate Pull Requests. I think it is specially useful in case there are errors because some similar packages are interdependent. 
The tests are passing so it is not so important in this case but for next time, it is better if we stick to our own guidelines :P

PR #453, opened 2023-05-24
Update Eslint dependencies 

ghost:
Since the new Eslint 3.37 and further version `property` dot should be on the same line what impossible in `<template>`.
Example:
`:label="messages.getUnescaped( 'wikibaselexeme-newlexeme-language' )"`
 :arrow_down_small: 
`:label="messages`
`.getUnescaped( 'wikibaselexeme-newlexeme-language' )"`

Solutions:
1. Possibly, it's Eslint bug and we should create a bug report;
2. Refactor our code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
 				"axe-core": "^4.7.1",
 				"cypress": "^12.13.0",
 				"cypress-axe": "^1.4.0",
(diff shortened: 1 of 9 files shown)


lucaswerkmeister:
I think we can close this in favor of #482 (which still has issues but is more recent).

PR #517, opened 2023-09-12
WIP: Add anon edit warning link parameters

micgro42:
TODO: figure out how to inject it properly

Bug: T330550


src/plugins/AccountLinkerPlugin/AccountLinker.ts
```diff
@@ -0,0 +1,21 @@
+import { inject, InjectionKey } from 'vue';
+
+export default interface AccountLinker {
```
micgro42, line 3: I think my main question is: "How to name this thing?" Better ideas welcome.

micgro42:
Note for review: this change is expected to work both with the old and the new `wikibase-anonymouseditwarning` message.

src/plugins/AccountLinkerPlugin/AccountLinker.ts
lucaswerkmeister: Maybe `AuthenticationLinker`?

I think “account” isn’t the usual term in a MediaWiki context, but `UserLinker` would probably be more confusing (that would link to someone else’s user page instead).

lucaswerkmeister (approved):
Tested with https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/955691/5 as the new version of the message.

src/plugins/AccountLinkerPlugin/AccountLinker.ts
micgro42: Good idea, thanks!

lucaswerkmeister (approved):

PR #518, opened 2023-09-20
Improve the required marker for fields

micgro42:
This moves the required marker closer to the actual label to make the connection more obvious. Also it makes it bigger.

The style blocks for overriding the gaps must not be scoped, because it otherwise the styles do not match anything.

The `line-height: 0;` has been added because otherwise the size of the asterisk affects the whole label and thus also the overall layout.

Bug: T322683

### DIFF
--- a/src/components/RequiredAsterisk.vue
+++ b/src/components/RequiredAsterisk.vue
@@ -24,6 +24,9 @@ export default {
 @import "@wmde/wikit-tokens/variables";
 
 .wbl-snl-required-asterisk {
+	font-size: $font-size-xxlarge;
+	line-height: 0;
+
 	&:not(:first-child) {
 		margin-inline-start: $dimension-spacing-small;
 	}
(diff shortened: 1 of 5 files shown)


src/components/RequiredAsterisk.vue
```diff
@@ -24,6 +24,8 @@ export default {
 @import "@wmde/wikit-tokens/variables";
 
 .wbl-snl-required-asterisk {
+	font-size: $font-size-xxlarge;
```
lucaswerkmeister, line 27: I think this should probably also add
```css
line-height: 0;
```
otherwise it significantly bumps the line height of the whole label, which affects the layout.

lucaswerkmeister (changes requested):
(Minor comment on the commit message: “because it otherwise” → “because otherwise”)

src/components/RequiredAsterisk.vue
lucaswerkmeister: Before/after:
image

micgro42: That looks much better indeed. I also like that now the asterisk has a bit of a "superscript" feel to it.

PR #533, opened 2023-10-06
Revert "Mostly downgrade required Node version from 16 to 14"

lucaswerkmeister:
The motivation at the time was apparently that [Fresh][1] only supported up to Node 14, but nowadays it supports up to 20, and 18 is the default. Meanwhile, some CI errors on #519 indicate that dependencies such as @vue/language-core are starting to require syntax like `??=` that’s not available in Node 14. Let’s bump the Node requirement to 16, the version we use in WikibaseLexeme CI.

This reverts commit db8ae7d515916b391d4624b2d5de0b60b14f7d54 (#151).

[1]: https://github.com/wikimedia/fresh

Bug: T348108

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -10,12 +10,7 @@ jobs:
             - name: Setup Node.js
               uses: actions/setup-node@v3
               with:
-                  node-version: "14"
-
-            - name: Setup NPM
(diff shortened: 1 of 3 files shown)


micgro42 (approved):
I searched the repository and this catches all the places where we set the node version to 14.

PR #535, opened 2023-10-09
Simplify createComponent() types in tests

lucaswerkmeister:
The `...config` in particular was causing type issues when upgrading @vue/test-utils in #492; and while that pull request is also blocked for other reasons, I think we can make our lives easier by simplifying these types. Callers of this method (which is local to this one test file) only pass in `props` and/or `global.provide`, so we can just specify the types for those members instead of full `ComponentOptions`; the `Props` type from `LemmaInput` is just duplicated, since I think that’s simpler than trying to get it back out of the full/exported `LemmaInput` type again.

Bug: T348108


lucaswerkmeister:
(Purely optional PR.)

micgro42 (approved):

PR #537, opened 2023-10-09
Document more ignored dependencies in README.md

lucaswerkmeister:
Based on #512, T341552 and #492.

Bug: T348108


micgro42 (approved):
Looking through https://www.npmjs.com/package/vue?activeTab=versions, it is surprising how many downloads the mediawiki-shipped version 3.2.37 still has compared to many of the 3.2.* versions that follow it.

PR #538, opened 2023-10-09
Update misc dev dependencies

lucaswerkmeister:
Bug: T348108


micgro42 (approved):
I looked at the [changelog for lint-staged](https://github.com/okonet/lint-staged/blob/master/CHANGELOG.md) and that major version bump seems to come from dropping Node 14 support and otherwise looks unremarkable.

PR #539, opened 2023-10-09
Update Vite-related packages

lucaswerkmeister:
As far as I can tell, this doesn’t seem to affect the compatibility of the generated code: while the output changes a bit, its size is almost identical; furthermore, the [Vite changelog](https://github.com/vitejs/vite/blob/v4.4.11/packages/vite/CHANGELOG.md) mentions that the nullish coalescing operator isn’t transpiled anymore since Vite 4, but we use such an operator in `actions.ts` and the corresponding build code can be found by searching for `.language.display.label`, and it doesn’t appear to have changed at all – so I think our build config is sufficiently specific that the change to Vite’s defaults doesn’t actually change the target version.

Bug: T348108

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
 				"@types/lodash": "^4.14.199",
 				"@typescript-eslint/eslint-plugin": "^5.61.0",
 				"@typescript-eslint/parser": "^5.61.0",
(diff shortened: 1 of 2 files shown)


micgro42 (approved):
I checked that the nullish coalescing operator (`??`) doesn't show up in the actual build output and it indeed does not.

PR #541, opened 2023-10-10
Update @rushstack/eslint-patch

lucaswerkmeister:
Bug: T348108


micgro42 (approved):
https://github.com/microsoft/rushstack/blob/main/eslint/eslint-patch/CHANGELOG.md is fine.

side note: I think there might also be some other projects of our that would benefit from us using this.

PR #543, opened 2023-10-11
Update more ESLint-related packages

lucaswerkmeister:
```sh
npm i @vue/eslint-config-typescript@latest
npm i {@typescript-eslint/{eslint-plugin,parser},@wmde/eslint-config-wikimedia-typescript,eslint-config-wikimedia}@latest
```

(Yes, it has to be two separate commands to work. I don’t know why and I don’t really care either.)

Bug: T348108


micgro42 (approved):
I was checking if the switch in `eslint-config-wikimedia` in 0.25.0 to es6 should make us update our eslint config, but we don't seem to mention es6 there at all, so this should be fine.

PR #545, opened 2023-10-23
Compare language codes case-insensitive

micgro42:
When switching to getting the language codes for a language from the IETF-code property in T348923, we also get a lot of composite codes for language variants that are partially uppercased, like en-GB or pt-BR. However, in our data, they are stored as completely lowercase strings. Thus, without this change, the user still has to manually, yet needlessly, select the valid language code from the dropdown.

TODO:
- [ ] this also has to work when _saving_ the statement, where the API also requires a fully lowercase language code

### DIFF
--- a/src/data-access/LanguageCodesProvider.ts
+++ b/src/data-access/LanguageCodesProvider.ts
@@ -18,6 +18,6 @@ export class MapLanguageCodesProvider implements LanguageCodesProvider {
 	}
 
 	public isValid( langCode: string ): boolean {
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
Hm, but I assume this means we send the language code as `en-GB` to the API when saving the lexeme… does that work? (I haven’t tried it yet.)

micgro42:
mh, good point. I'll try it locally

micgro42:
This needs a different approach. Probably in `MwApiLangCodeRetriever`?

PR #546, opened 2023-10-23
DevLangCodeRetriever: Adjust property-id used for language code

micgro42:
This was changed in production to user the broader coverage provided by IETF code, see T348923.

Bug: T348923

### DIFF
--- a/src/data-access/DevLangCodeRetriever.ts
+++ b/src/data-access/DevLangCodeRetriever.ts
@@ -6,7 +6,7 @@ export default class DevLangCodeRetriever implements LangCodeRetriever {
 		const response = await fetch( 'https://www.wikidata.org/w/api.php?' + new URLSearchParams( {
 			action: 'wbgetclaims',
 			entity: itemId,
-			property: 'P218',
+			property: 'P305',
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Hm, this doesn’t seem to be working properly for me:
image
It’s making the right API request, and the API returns `en-GB`, yet it neither accepts the language code nor complains that it’s invalid. It’s acting as if there was no statement… any idea why?

lucaswerkmeister:
Aha, I’m pretty sure `return processWbGetClaimsResponse( response, 'P218' );` also needs to use the new property ID :)

lucaswerkmeister:
Yeah, with that it seems to be working (“This Item has an unrecognized language code. Please select one below.”).

lucaswerkmeister (approved):

micgro42:
ok, with that cypress fixup it should be fine now

lucaswerkmeister (approved):
(`MwApiLangCodeRetriever.test.ts` also uses P218 in some test data, but I think that’s fully unrelated and can be updated separately.)

PR #548, opened 2023-10-30
Lowercase language code from item statement

lucaswerkmeister:
We now use the property “IETF language tag” for the automatic language code on Wikidata; the subtag of these tags is case-insensitive, but some capitalization is recommended.<sup>[1][1], [2][2]</sup> For MediaWiki or Wikidata, on the other hand, the subtags generally need to be all-lowercase (except our special mis-x-Qid codes, see T317863). Reconcile this by lowercasing all language codes from items in the store action.

[1]: https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags
[2]: https://en.wikipedia.org/w/index.php?title=IETF_language_tag&oldid=1181524280#Syntax_of_language_tags

Bug: T349652

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -221,6 +221,9 @@ export default function createActions(
 			{ commit }: RootContext,
 			languageCode: string | undefined | null | false,
 		) {
+			if ( typeof languageCode === 'string' ) {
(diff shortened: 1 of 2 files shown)


micgro42 (approved):
Looks good, thank you!

PR #570, opened 2023-11-14
Split comment across two lines

lucaswerkmeister:
To prevent an eslint warning about the line length that’s showing up on every pull request (under “Unchanged files with check annotations”) and starting to get annoying.

Bug: T349652
Bug: T350664


micgro42 (approved):

micgro42:
I wonder if we should change eslint to error on warnings. That way we have to fix these things right away.

lucaswerkmeister:
Yeah, maybe. (It’s a bit annoying that you can’t do this via eslintrc, you have to add it to the command line, AFAIK.)

PR #578, opened 2023-11-15
Update README.md for November 2023 chores

lucaswerkmeister:


README.md
```diff
@@ -111,6 +111,9 @@ The following dependencies should be ignored:
 - Vue Test Utils (`@vue/test-utils`):
   Seems to require a newer Vue.js version (see [#492](https://github.com/wmde/new-lexeme-special-page/pull/492)),
   though it’s worth checking whether that’s still the case in newer VTU versions.
+- lint-staged:
+  [v15.0.0](https://github.com/lint-staged/lint-staged/releases/tag/v15.0.0 drops compatibility with Node 16,
+  which we still use in CI (see also “Prerequisites” above).
```
micgro42, line 116: Mh, why are we still using that in CI? As I understand it, Node 18 and Node 20 images already exist: [⚓ T331181 Create WMF CI images and jobs for Node.js 18](https://phabricator.wikimedia.org/T331181). We should probably at least have a (maybe stalled) task to upgrade this Node 18 or maybe directly to Node 20. And mention/link the task here.

README.md
```diff
@@ -130,6 +133,10 @@ For example, this can be one commit for each of:
 
 Make sure that all checks still pass for every commit.
 
+When using dependabot to update dependencies (where there is nothing to group),
+you can trigger a new dependabot run (to potentially open additional pull requests)
+at [Insights > Dependency graph > Dependabot](https://github.com/wmde/new-lexeme-special-page/network/updates) > Recent update jobs > Check for updates.
```
micgro42, line 138: Nice, I wasn't aware that this is possible! However, from reading this paragraph, I'm not yet sure why I would want to do that. Can you elaborate?

README.md
micgro42: Ah, this would probably actually be waiting on [⚓ T331180 Upgrade all CI jobs for WMF-deployed projects from Node 16 to Node 18](https://phabricator.wikimedia.org/T331180) instead.

lucaswerkmeister: Yeah, the images exist (Fresh also uses them) but they’re not used in CI yet.

I guess `lint-staged` in particular doesn’t need to run in CI, but if we upgrade it then we get into a weird situation where you need Node 18 to develop locally, but also the build still has to stay compatible with Node 16 (so we can check the dist diff in CI)…

README.md
lucaswerkmeister: I guess it depends on the workflow you use. This time, I tried to do as many updates as seemed reasonable / justifiable using dependabot, to minimize the number of review round-trips (which take much longer than me just interacting with dependabot), especially since those can’t really be parallelized (I could upload several PRs, but merging one would probably invalidate the other and rebasing that would discard the review, so I’d still have to wait again, because GitHub is very inflexible about code review compared to Gerrit). But this kind of contradicts the process above where we want to “keep the number of commits … manageable” :/

README.md
micgro42: Makes sense. Then linking T331180 in this section would be good enough for me. Then we can easily check during future chores if we can move forward with this or still have to wait.

README.md
micgro42: Ah makes sense, maybe we could add a sentence somewhere about "to make sure the existing PRs are fully up-to-date) or something similar?

README.md
lucaswerkmeister: Done.

README.md
lucaswerkmeister: I rephrased it a bit.

README.md
```diff
@@ -111,6 +111,9 @@ The following dependencies should be ignored:
 - Vue Test Utils (`@vue/test-utils`):
   Seems to require a newer Vue.js version (see [#492](https://github.com/wmde/new-lexeme-special-page/pull/492)),
   though it’s worth checking whether that’s still the case in newer VTU versions.
+- lint-staged:
+  [v15.0.0](https://github.com/lint-staged/lint-staged/releases/tag/v15.0.0 drops compatibility with Node 16,
```
micgro42, line 115: This link is now missing a closing brace.

README.md
micgro42: Better, thanks!

micgro42 (changes requested):
Looks good, just one small syntax error remaining, then it's good to go!

README.md
lucaswerkmeister: thanks, done

micgro42 (approved):

PR #581, opened 2023-11-30
Remove TypeScript and Vue Test Utils notes from README.md

lucaswerkmeister:
Now that we upgraded from Vue 3.2 to 3.3, TypeScript is no longer *known* to be blocked, so the next round of chores should attempt to upgrade it. If there are any problems, we can add it back to the list.

Whether Vue Test Utils is unblocked is even less clear to me, but let’s still remove it and potentially put it back if needed.

Bug: T352388


micgro42:
This might also apply to the Vue Test Utils item in that list. What do you think?

micgro42 (approved):

PR #594, opened 2023-12-04
Update lint-staged and prettier

codders:
```
npm i lint-staged@15.2.0
npm i prettier@3.1.0
```

`lint-staged` after version v15 has a minimum node version of 18.12.0 - updated README accordingly.
`prettier@3` adds support for Vue3 generic components.

Bug: T352521


lucaswerkmeister:
Both of these are actually blocked according to the README:

> - Prettier:
>   [Jest 29.65.2](https://github.com/jestjs/jest/blob/main/CHANGELOG.md#2962) is not compatible with Prettier 3.
>   This might be resolved with the next release ([Jest#14566](https://github.com/jestjs/jest/pull/14566)),
>   so be sure to check whether a newer Jest version is available that includes Prettier 3 compatibility.
> - lint-staged:
>   [v15.0.0](https://github.com/lint-staged/lint-staged/releases/tag/v15.0.0) drops compatibility with Node 16,
>   which we still use in CI (see also “Prerequisites” above; Node 18 upgrade pending in [T331180](https://phabricator.wikimedia.org/T331180)).

I haven’t checked whether there’s a newer Jest release available by now, so maybe Prettier can be unblocked if we upgrade Jest too. But lint-staged definitely needs to stay on v14 for the time being.

codders:
Okay - thanks for the tips!

PR #596, opened 2023-12-05
Update Vue Test Utils

lucaswerkmeister:
```bash
npm r @vue/test-utils && npm i -D @vue/test-utils && npm dedupe
```

npm can’t update the dependency “in place”, but removing it first seems to resolve the peer dependency conflicts. npm dedupe then helps with some, but not all, errors at runtime.

SpellingVariantInput needs a slight change to make TypeScript pass.

Bug: T352749


codders:
Superceded by #670

PR #598, opened 2023-12-27
Change the Anonymous Edit Warning to add TempUser support

codders:
In the case that the Mediawiki has temporary users enabled, the user's IP address will no longer be displayed in the case of an anonymous edit. Update AnonymousEditWarning to reflect this.

Bug: T352016


codders:
Is made from the wrong repository (i.e. my fork) and therefore fails CI. I will close this PR and open a new one.

PR #610, opened 2024-01-02
Add the reason why we cannot update the Node.js version to the README

micgro42:
I forget about every two months why exactly we cannot upgrade to a newer node version just in the GitHub actions/workflows. Better write that down right where one would expect it.


codders (approved):
Thanks - that's helpful!

PR #611, opened 2024-01-03
Change the Anonymous Edit Warning to add TempUser support

codders:
In the case that the Mediawiki has temporary users enabled, the user's IP address will no longer be displayed in the case of an anonymous edit. Update AnonymousEditWarning to reflect this.

Bug: T352016


micgro42 (approved):
Looks good to me, thanks!

PR #635, opened 2024-02-02
Upgrade dependencies for node 18

codders:
We were stuck for a while on node 16, but now we have node 18 we can upgrade a few of the libraries:

 - @namics/stylelint-bem
   - drop support for node < 18.12 and stylelint < 14
 - lint-staged
   - drop support for node < 18.12
   - various minor fixes
 - vite
   - drop support for node < 18
   - migrate to Rollup v4
   - numerous bugfixes and performance improvements
 - @vitejs/plugin-vue
   - drop support for node < 18
   - drop support for vite < 5
   - minor fixes

Bug: T356273


lucaswerkmeister (approved):
Thanks for the detailed commit message! (And apparently the build output shrinks by 54 bytes, nice.)

lucaswerkmeister (approved):

PR #641, opened 2024-02-29
Fix redirecting to URL without temporary user

lucaswerkmeister:
`mw.util.getUrl()` can return a relative URL, which is illegal to pass directly into `new URL()` unless you also specify the base. Do that.

Directly using `window.location.href` as the base feels a bit ugly, but I don’t think injecting it as a separate parameter of the service would be nicer.

Bug: T358754


lucaswerkmeister:
I’ll see if I can also add a test for this.

codders (approved):
Looks great - thanks for taking care of this!

PR #731, opened 2024-08-12
Update semver to resolve npm audit warning

codders:
Resolve REDoS issue with Semver, which is included as a dependency of libraries we are using.

https://github.com/advisories/GHSA-c2qf-rxjj-qqgw

Bug: T370126


lucaswerkmeister (dismissed):
I’d be willing to bet quite a bit of money that there’s no possible way we’re affected by this vulnerability (when would we ever pass untrusted user input to semver?), but such are the joys of npm.

PR #745, opened 2024-09-09
POC: Start using Codex

lucaswerkmeister:
I thought the `TextInput` would be an easy component to start with. I was wrong, but hey, at least this mostly works. I also left some TODOs in the code as a treat.

Bug: T369505


lucaswerkmeister:
Superseded by #746, I think.

PR #747, opened 2024-09-10
POC: Use Codex in NewLexemeForm

lucaswerkmeister:
Using the Button component from Codex instead of Wikit is straightforward – we just need to rename a few props.

Using the variables is a bit more involved. Not only do we need replacement variables from Codex that aren’t always clear, but the small-text mixin used for the copyright notice also relies on the Wikit variables. So for now, this change just keeps importing the Wikit variables and lets Codex override them wherever they both use the same name.

Bug: T369505

----

Based on #746.


lucaswerkmeister:
Oh, oops, closed the wrong PR I think

PR #750, opened 2024-09-19
Use Codex tokens and custom spacing variables

codders:
Migrate away from Wikit tokens to Codex SASS variables where they are available. In some cases we use custom CSS variables for values from Wikit whose Codex equivalents are missing.

Bug: T369510
Bug: T369508
Bug: T369509


yerdua (approved):

codders:
superseded by #762

PR #760, opened 2024-09-19
Update css selector target for button selection

codders:


lucaswerkmeister:
I’ve adopted the fix in #761, thank you!

PR #762, opened 2024-09-20
Use Codex tokens and custom spacing variables

codders:
Migrate away from Wikit tokens to Codex SASS variables where they are available. In some cases we use custom CSS variables for values from Wikit whose Codex equivalents are missing.

Bug: [T369510](https://phabricator.wikimedia.org/T369510)
Bug: [T369508](https://phabricator.wikimedia.org/T369508)
Bug: [T369509](https://phabricator.wikimedia.org/T369509)

### DIFF
--- a/src/components/LanguageInput.vue
+++ b/src/components/LanguageInput.vue
@@ -79,11 +79,11 @@ export default {
 </template>
 
 <style lang="scss">
-@import '@wmde/wikit-tokens/variables';
+@import '@wikimedia/codex-design-tokens/theme-wikimedia-ui';
 
 /* stylelint-disable selector-class-pattern */
 .wbl-snl-language-lookup .wikit .wikit-Lookup__label-wrapper {
-	gap: $dimension-spacing-xsmall;
+	gap: $spacing-50;
 }
 /* stylelint-enable selector-class-pattern */
 
--- a/src/components/NewLexemeForm.vue
+++ b/src/components/NewLexemeForm.vue
@@ -191,30 +191,31 @@ export default {
 </template>
 
 <style scoped lang="scss">
-@import '@wmde/wikit-tokens/variables';
-@import '@wmde/wikit-vue-components/src/styles/mixins/Typography';
+@import '@/styles/custom-variables.css';
+@import '@wikimedia/codex-design-tokens/theme-wikimedia-ui';
 
 .wbl-snl-form {
 	& > * + * {
-		margin-top: $dimension-layout-xsmall;
+		margin-top: var( --dimension-layout-xsmall );
 	}
 
 	// Box model
-	padding: $dimension-layout-small;
+	padding: var( --dimension-layout-small );
 
 	// Border
 	border-style: $border-style-base;
-	border-width: $border-width-thin;
+	border-width: $border-width-base;
 	border-radius: $border-radius-base;
-	border-color: $border-color-base-subtle;
+	border-color: $border-color-muted;
 }
 
 .wbl-snl-copyright {
-	@include small-text;
-
-	font-size: 0.8125rem;
-	font-style: italic;
-	font-synthesis: none;
+	/* codex Body S */
+	font-family: $font-family-system-sans;
+	font-size: $font-size-small;
+	font-weight: $font-weight-normal;
+	line-height: $line-height-medium;
+	color: $color-base;
 	margin-bottom: 0;
 }
 
--- a/src/components/RequiredAsterisk.vue
+++ b/src/components/RequiredAsterisk.vue
@@ -21,18 +21,12 @@ export default {
 </template>
 
 <style lang="scss" scoped>
-@import '@wmde/wikit-tokens/variables';
+@import '../styles/custom-variables.css';
+@import '@wikimedia/codex-design-tokens/theme-wikimedia-ui';
 
 .wbl-snl-required-asterisk {
-	font-size: $font-size-xxlarge;
+	font-size: $font-size-x-large;
 	line-height: 0;
-
-	&:not( :first-child ) {
-		margin-inline-start: $dimension-spacing-small;
-	}
-
-	&:not( :last-child ) {
-		margin-inline-end: $dimension-spacing-small;
-	}
+	margin-inline-start: var( --dimension-spacing-xsmall );
 }
 </style>
--- a/src/components/SearchExisting.vue
+++ b/src/components/SearchExisting.vue
@@ -24,13 +24,19 @@ const searchMessage = computed( () => messages.get(
 </template>
 
 <style lang="scss" scoped>
-@import '@wmde/wikit-tokens/variables';
-@import '@wmde/wikit-vue-components/src/styles/mixins/Typography';
+@import '@/styles/custom-variables.css';
+@import '@wikimedia/codex-design-tokens/theme-wikimedia-ui';
 
 .wbl-snl-search-existing {
-	@include body;
+	/* font Codex Body */
+	font-family: $font-family-system-sans;
+	font-size: $font-size-small;
+	font-weight: $font-weight-normal;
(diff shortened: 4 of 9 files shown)


lucaswerkmeister:
Is there a reason to make `--dimension-layout-xsmall` a CSS variable? Looking at this change overall, to me it feels like it would make more sense to make it a Sass variable like all the other tokens. (This would also fix the minor issue that the current version includes the variable styles  three times in the build output as part of different components.)

src/components/LanguageInput.vue
```diff
@@ -79,11 +79,11 @@ export default {
 </template>
 
 <style lang="scss">
-@import '@wmde/wikit-tokens/variables';
+@import '@wikimedia/codex-design-tokens/theme-wikimedia-ui';
 
 /* stylelint-disable selector-class-pattern */
 .wbl-snl-language-lookup .wikit .wikit-Lookup__label-wrapper {
-	gap: $dimension-spacing-xsmall;
+	gap: $spacing-50;
```
lucaswerkmeister, line 86: This increases the gap from `.25em` (normally = `4px`) to `8px`. According to T369510 that’s intended. Likewise in `LemmaInput`, `LexicalCategoryInput`, `RequiredAsterisk`, `SpellingVariantInput`.

src/components/NewLexemeForm.vue
```diff
@@ -190,31 +190,31 @@ export default {
...
 
 .wbl-snl-form {
 	& > * + * {
-		margin-top: $dimension-layout-xsmall;
+		margin-top: var( --dimension-layout-xsmall );
 	}
 
 	// Box model
-	padding: $dimension-layout-small;
+	padding: var( --dimension-layout-xsmall );
```
lucaswerkmeister, line 202: This is supposed to be `--dimension-layout-small`, not **x**small, according to T369510. (Assuming we want to preserve the previous value, that variable should be set to `1.5rem`.)

src/components/NewLexemeForm.vue
```diff
@@ -190,31 +190,31 @@ export default {
...
 
 	// Border
 	border-style: $border-style-base;
-	border-width: $border-width-thin;
+	border-width: $border-width-base;
 	border-radius: $border-radius-base;
-	border-color: $border-color-base-subtle;
+	border-color: $border-color-muted;
```
lucaswerkmeister, line 208: Changes from `#c8ccd1` to `#eaecf0`; intentional according to T369508, I guess.

src/components/NewLexemeForm.vue
```diff
@@ -190,31 +190,31 @@ export default {
...
 .wbl-snl-copyright {
-	@include small-text;
-
-	font-size: 0.8125rem;
-	font-style: italic;
-	font-synthesis: none;
-	margin-bottom: 0;
+	/* codex Body S */
+	font-family: $font-family-system-sans;
+	font-size: $font-size-small;
```
lucaswerkmeister, line 214: Slight change from `.8125rem` to `.875rem` is expected according to T369509.

src/components/NewLexemeForm.vue
```diff
@@ -190,31 +190,31 @@ export default {
...
+	border-color: $border-color-muted;
 }
 
 .wbl-snl-copyright {
-	@include small-text;
-
-	font-size: 0.8125rem;
-	font-style: italic;
-	font-synthesis: none;
-	margin-bottom: 0;
```
lucaswerkmeister, line 217: AFAICT none of the tasks mentions removing this margin override? (Though it doesn’t seem to make a difference either way because both the button and the error message have a larger top margin anyway, which collapses with the copyright notice’s bottom margin whether it’s `0.5em` or `0`.)

src/components/RequiredAsterisk.vue
```diff
@@ -21,18 +21,18 @@ export default {
 </template>
 
 <style lang="scss" scoped>
-@import '@wmde/wikit-tokens/variables';
+@import '@wikimedia/codex-design-tokens/theme-wikimedia-ui';
 
 .wbl-snl-required-asterisk {
-	font-size: $font-size-xxlarge;
+	font-size: $font-size-xx-large;
```
lucaswerkmeister, line 27: This changes the font size from `1.5em` to `1.5rem`, which actually makes a difference (at least in legacy Vector) – I don’t think this is expected by T369509, which says that “that same token […] matches Codex already”?

src/components/SearchExisting.vue
```diff
@@ -24,13 +24,19 @@ const searchMessage = computed( () => messages.get(
...
-@import '@wmde/wikit-tokens/variables';
-@import '@wmde/wikit-vue-components/src/styles/mixins/Typography';
+@import '@/styles/custom-variables.css';
+@import '@wikimedia/codex-design-tokens/theme-wikimedia-ui';
 
 .wbl-snl-search-existing {
-	@include body;
+	/* font Codex Body */
+	font-family: $font-family-system-sans;
+	font-size: $font-size-medium;
```
lucaswerkmeister, line 33: This changes the font size from `1em` to `1rem`, which on my wiki corresponds to `14px` and `16px` respectively – according to T369509 we should keep `14px`.

src/components/SearchExisting.vue
```diff
@@ -24,13 +24,19 @@ const searchMessage = computed( () => messages.get(
...
+	font-weight: $font-weight-normal;
+	line-height: $line-height-medium;
```
lucaswerkmeister, line 35: This changes from `1.5` to `1.6` – not explicitly mentioned in T369509, but expected, I guess.

lucaswerkmeister (commented):
Tried to go through the changes and figure out which task they correspond to and whether they’re right or not. Many of the comments are just acknowledgements, but I think some represent issues.

src/components/NewLexemeForm.vue
codders: Added this back - thanks for catching this.

src/components/NewLexemeForm.vue
codders: Fixed this - thanks for the catch!

codders:
> Is there a reason to make `--dimension-layout-xsmall` a CSS variable? Looking at this change overall, to me it feels like it would make more sense to make it a Sass variable like all the other tokens. (This would also fix the minor issue that the current version includes the variable styles three times in the build output as part of different components.)

I'm not super opinionated. I think the intention was to have a CSS file at the end of the migration that we could try an upstream to Codex to have them add the variables to the CSS, but we could also keep the missing variables in Sass in our repo for now.

lucaswerkmeister:
> I'm not super opinionated. I think the intention was to have a CSS file at the end of the migration that we could try an upstream to Codex to have them add the variables to the CSS, but we could also keep the missing variables in Sass in our repo for now.

I’d prefer to make them Sass variables for now and then potentially reevaluate that at the end of T375370, but I’m also okay with keeping them as they are.

(The other questions are now blocked on feedback from UX IIUC.)

codders:
Updated per UX feedback - should be good to go now.

src/components/RequiredAsterisk.vue
lucaswerkmeister: Now `1.25rem`, which comes out to `20px` rather than `21px` (smaller change).

src/components/SearchExisting.vue
lucaswerkmeister: Fixed.

lucaswerkmeister:
There’s currently too much space before the lexeme language / lexical category asterisks, but I think that’s just a consequence of #773 being blocked – the asterisks currently have both the gap on the surrounding label and the margin that’s supposed to replace the gap, if I understand correctly.
image

lucaswerkmeister (approved):

PR #765, opened 2024-09-23
Use Codex Field + TextInput instead of Wikit TextInput

lucaswerkmeister:
Change the lemma to not use Wikit anymore. In Codex, this corresponds to two components, so use both.

The spacing of the “required” asterisk is done differently – it’s no longer in a flex container, so establish the spacing using margin instead. The asterisk is now also part of the actual <label> element, which isn’t ideal but I think can’t be avoided within Codex. (Note that Codex doesn’t want us to indicate required fields at all; see also T374432.)

Bug: T370054

----

The second commit is optional, if you don’t like it I can take it out again:

----

Restructure error computation

Since Codex wants the status and messages as separate props, but we compute them at the same time, let’s have the computed return an object with both props.

### DIFF
--- a/src/components/LemmaInput.vue
+++ b/src/components/LemmaInput.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { TextInput } from '@wmde/wikit-vue-components';
+import { CdxField, CdxTextInput } from '@wikimedia/codex';
 import RequiredAsterisk from '@/components/RequiredAsterisk.vue';
 import { useMessages } from '@/plugins/MessagesPlugin/Messages';
 import { useConfig } from '@/plugins/ConfigPlugin/Config';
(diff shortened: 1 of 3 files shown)


src/components/LemmaInput.vue
```diff
@@ -51,31 +61,33 @@ export default {
...
+		<!-- eslint-enable -->
+		<template #label>
+			{{ messages.getUnescaped( 'wikibaselexeme-newlexeme-lemma' ) }}<required-asterisk />
 		</template>
-	</text-input>
+	</cdx-field>
 </template>
 
 <style lang="scss">
 @import '@wmde/wikit-tokens/variables';
```
codders, line 88: Should we not be removing this import? And replacing the SASS variable below with a CSS variable?

lucaswerkmeister: In principle yes, but I thought this would be part of a separate task (T369510 if I’m not mistaken) and not included in this PR.

codders: eh. fair enough. I guess it will be some weeks before UX responds on that PR, so I'll end up rebasing it anyway.

codders (approved):
lgtm +1

PR #767, opened 2024-09-27
Remove wikit link component

yerdua:
Replaces the Wikit Link component with a plain link, which gets codex styling through the default stylesheet.

Bug: T370053

### DIFF
--- a/src/components/SpellingVariantInput.vue
+++ b/src/components/SpellingVariantInput.vue
@@ -2,7 +2,6 @@
 import { ref, computed } from 'vue';
 import escapeRegExp from 'lodash/escapeRegExp';
 import WikitLookup from './WikitLookup';
-import { Link as WikitLink } from '@wmde/wikit-vue-components';
 import RequiredAsterisk from '@/components/RequiredAsterisk.vue';
 import { useMessages } from '@/plugins/MessagesPlugin/Messages';
 import { useLanguageCodesProvider } from '@/plugins/LanguageCodesProviderPlugin/LanguageCodesProvider';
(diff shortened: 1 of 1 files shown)


yerdua:
> https://d6416fde732a76091ace017292d52e9ced9e--new-lexeme-special-page.netlify.app

When viewing this in the context of a mediawiki installation, all of the links are styled correctly with Codex, using the default css. When viewing in ^that standalone preview, they are not. I'm assuming this is fine?

codders (approved):
From a code point of view, this looks fine. But it doesn't seem to fulfill the requirement to use / reproduce `cdx-mixin-link`. If you're satisfied that the `<a>` tag gets all the styles that `cdx-mixin-link` would apply, then I guess we just merge this and see if it makes it through UX review.

I'll approve this and let you decide whether to merge.

yerdua:
the codex-mixin-link is applied in Vector's [resources/skins.vector.styles/links.less](https://gerrit.wikimedia.org/g/mediawiki/skins/Vector/+/d605035b9740e50e7aab1ab50ced6a2fe20d19b5/resources/skins.vector.styles/links.less#2) and in MinervaNeue's [resources/skins.minerva.base.styles/content/links.less](https://gerrit.wikimedia.org/g/mediawiki/skins/MinervaNeue/+/a0a50482e9d3c4cb07cefd5bfa6db0daf1b3673d/resources/skins.minerva.base.styles/content/links.less#28).

At least for these skins, I'm convinced this is enough to fulfill the requirements. And if I'm wrong, this is at least a step on the way. So, I'll merge it.

PR #776, opened 2024-10-07
Submit form using click handler

lucaswerkmeister:
Putting the event handler on the button seems more intuitive to me, and works around T376622 (a Codex button doesn’t submit the form when activated using the keyboard). Note that a “click” handler on a button also fires when it’s activated using the keyboard (this is standard browser behavior but still works on a Codex button).

As far as I’m aware, there was no special reason to have the event handler on the form before – the event handler was added in e35cdf36d1, replacing a different attribute on the `<form>`, so it probably just seemed natural to put the event handler on the `<form>` at the time.

Bug: T370052
Bug: T376622


codders (approved):
Looks good to me - thanks!

PR #777, opened 2024-10-11
Fix Vue version

lucaswerkmeister:
See individual commit messages. (Reverts #774.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,9 @@ updates:
       schedule:
           interval: monthly
       open-pull-requests-limit: 15
+      ignore:
+          # use the same version of Vue and Vuex as in MediaWiki core
+          - dependency-name: "vue"
+          - dependency-name: "@vue/compiler-dom"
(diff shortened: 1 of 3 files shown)


yerdua (approved):

lucaswerkmeister:
Updated after conflicts (I ended up just recreating the first commit from scratch).

lucaswerkmeister:
Wondering if we should also pin the versions in `package.json` as `3.4.27` rather than `^3.4.27`…

yerdua:
> Wondering if we should also pin the versions in `package.json` as `3.4.27` rather than `^3.4.27`…

Makes sense to me. Is there any harm in doing so?

lucaswerkmeister:
Don’t think so, I just didn’t think of it yesterday and then thought of it today while idly staring at the terminal output from yesterday ^^ I’ll update the PR.

lucaswerkmeister:
Ended up putting it in a separate commit because I didn’t think it really fit into either of the existing ones.

yerdua (approved):

PR #800, opened 2024-11-13
Fix unreliable input on SpellingVariantInput

codders:
In the port from Wikit to Codex, the new implementation of `SpellingVariantInput` was working, but was dropping some inputs entered by users.

Remove the unused `search-input` attribute of the Codex Lookup component in `SpellingVariantInput` and wire the `input-value` property in the same way it has been connected in the `ItemLookup` implementation.

Bug: T379595


mariushoch (approved):
Also fixes the lexeme e2e tests :)

PR #801, opened 2024-11-15
Remove unnecessary `onOptionSelected` call

codders:
The code for `SpellingVariantInput` was making an unnecessary extra call to `onOptionSelected` when `onInput` was called, to satisfy assertions in the unit test suite that are very involved with the internals of the component eventing logic. This worked okay for codex 1.14.0, but is incompatible with the behaviour changes introduced in Icd4b7ed1e34bf5b2d62f705fdd8c6f4e0cf088d1.

Remove the unnecessary call, and remove the unnecessarily detailed unit tests.

Bug: T379595


lucaswerkmeister (approved):
LGTM; the behavior is apparently sufficiently covered by the integration tests (once we use the right Codex version, anyway).

PR #808, opened 2024-11-27
Enable bold labels in item lookup

lucaswerkmeister:
Per UX feedback.

Bug: T370057


codders (approved):
Looks good - thank you!

PR #833, opened 2025-01-21
Update Vue, Codex and related dependencies

mariushoch:
I needed to use `--legacy-peer-deps` here as npm seemingly doesn't allow installing/ updating a package and its peer dependencies (@wikimedia/codex which has a peer dependency on vue, thus they would need to be updated in a single transaction). This seems to be desired behaviour, per: npm/cli#7719 (comment)

`npm install --legacy-peer-deps vue@3.5.13 vue-tsc@2.2.0 vuex@4.1.0 @wikimedia/codex@1.20.0 @wikimedia/codex-design-tokens@1.20.0 @vue/server-renderer@3.5.13`

Also some, seemingly harmless, changes in tests where needed. 

Bug: T382330

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,12 +9,12 @@
 			"version": "0.0.1",
 			"license": "GPL-2.0-or-later",
 			"dependencies": {
-				"@wikimedia/codex": "^1.16.0",
-				"@wikimedia/codex-design-tokens": "^1.16.0",
+				"@wikimedia/codex": "^1.20.0",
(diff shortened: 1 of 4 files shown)


lucaswerkmeister:
IIUC all the Vue/Codex updates should be released now (MediaWiki core updated and Codex released a new version), so hopefully we can pin them again.

lucaswerkmeister (dismissed):

lucaswerkmeister (approved):

PR #836, opened 2025-03-21
Avoid compiling color constants to static values for dark-mode

codders:
SpecialNewLexeme is substantially built using Codex components, which automatically have correct colour variables and support for darkmode. In codex itself, components are styled using LESS (`<style lang="less">`) which allows variables in the stylesheets to refer to their dynamic values in the compiled code.

For components that are specific to SpecialNewLexeme, we have used SCSS (`<style lang="scss">`) and imported the codex variables via codex's CSS export. Referring to these values (e.g. `$color-base`) from SCSS evaluates the current value of the variable at compile time and hard-codes that into the generated Vue code.

Use the CSS variables (`var( --border-color-muted )`) to refer to codex colour values from SCSS so that the darkmode skin can override the colour values at runtime.

Bug: T385046


yerdua (approved):
looks good to me!

